PR #2, opened 2015-09-06
Fix the L

mdinger:
Silly mistake with an unhelpful error. This makes the example work now. Reminds me of https://github.com/rust-lang/rust/issues/13677

### DIFF
--- a/index.md
+++ b/index.md
@@ -58,7 +58,7 @@ use gtk::signal::Inhibit;
 
 fn main() {
     gtk::init().unwrap_or_else(|_| panic!("Failed to initialize GTK."));
-    let window = gtk::Window::new(gtk::WindowType::Toplevel).unwrap();
+    let window = gtk::Window::new(gtk::WindowType::TopLevel).unwrap();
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Thanks for your PR ! However, in the current master branch, this is `Toplevel` and not `TopLevel`. I'll try to update crates repos as soon as possible.

mdinger:
Ooooooh, I see.

gkoz:
Right, the breaking change. The updated crates would be accompanied by a blog post explaining that but it has bitten you in an unexpected way. I guess this PR might actually be good to apply now, while the crates are still at the current version.

GuillaumeGomez:
@gkoz: I'll push the master branches on crates tomorrow. I'll discover how the news system is working then ! :D

gkoz:
@mdinger would you reopen the PR please. It does make sense to make the example work with the current crate versions.

mdinger:
There you go

gkoz:
Thanks

PR #3, opened 2015-09-06
Example code improvement

GuillaumeGomez:
cc @gkoz

### DIFF
--- a/index.md
+++ b/index.md
@@ -57,7 +57,7 @@ A create a window, etc.
 use gtk::signal::Inhibit;
 
 fn main() {
-    gtk::init().unwrap_or_else(|_| panic!("Failed to initialize GTK."));
+    gtk::init().ok().expect("Failed to initialize GTK.");
     let window = gtk::Window::new(gtk::WindowType::TopLevel).unwrap();
     window.set_title("First GTK+ Program");
     window.set_default_size(350, 70);


gkoz:
I'm not enthusiastic about endorsing `.ok().expect(...)`...

GuillaumeGomez:
Hum... Any particular reason ?

gkoz:
It works here because our error type is `()` but generally is inferior to `Result::expect` and seems the opposite of idiomatic. People will use pieces of our examples without questioning them. Maybe we should replace the panic with `print!` and `return`.

GuillaumeGomez:
I'd prefer this actually. I'm not a big fan of `panic!` when it's not absolutely necessary.

gkoz:
We should start by updating the actual example(s) anyway and then bring the change here.

GuillaumeGomez:
Actually, it doesn't seem to be a priority but if you have nothing else waiting, then I let you update examples. ;)

gkoz:
I'm working on push some hand-generated docs after that maybe.

GuillaumeGomez:
As you wish, I'm starting to prepare to push on crates.io tomorrow.

gkoz:
I've found a little issue with the version handling and enums in cairo. Might want to hold off pusblishing crates a bit.

GuillaumeGomez:
Ok, I'm in standby. Just email me when the situation is unlocked please. ^^

gkoz:
I've just learnt more about the issue. Good news is, it's not a crates showstopper.

Turns out rustdoc doesn't get the same `cfg` directives `rustc` gets so it can't see any of the gated stuff. WAT

GuillaumeGomez:
DARK MAGIC !!!
Well, I work on crates tomorrow now. Witches will wait !

gkoz:
Looks like we've resolved this in a different way.

PR #4, opened 2015-09-07
Link to docs and gitter

gkoz:
This isn't a very scalable approach but works for the few links we currently have :p


GuillaumeGomez:
We'll see if we need more links later. ^^

PR #14, opened 2016-03-12
Tutorial basics

GuillaumeGomez:


GuillaumeGomez:
@gkoz: Here are the basis of the tutorial. I only wrote the gtk-rs specific chapters. I think it'll be enough for the release. Now remains to finish and improve them.

gkoz:
Site structure note: these should go into a subdirectory of `docs-src` and not try to be a blog.

_tutorial/part_1.md
```diff
@@ -0,0 +1,79 @@
...
+title: Tutorial - Part 1
+date: 2016-03-11 18:00:00 +0100
+categories: [front, tutorial]
+---
+
+Through these chapters, this tutorial intends to allow you to understand how to use gtk-rs' libraries. Let's start with the baics: building a basic application.
+
+### Basics
+
+Before going any further, let's start with the `cargo` file. Create a `Cargo.toml` like the following:
```
gkoz: I think a more idiomatic starting point would be `cargo new --bin` and then adding the dependency.

_tutorial/part_1.md
```diff
@@ -0,0 +1,79 @@
...
+authors = ["You Name <your@email.com>"]
+
+[dependencies]
+gtk = "0.0.7"
+
+[[bin]]
+name = "gtk_tuto"
+```
+
+So we created a project named "tuto" which generates a binary "gtk_tuto" and has both `gtk` as dependency. Now let's create a `gtk_tuto.rs` into a `src` directory file:
```
gkoz: Then you would edit `main.rs` created by cargo.

_tutorial/part_1.md
```diff
@@ -0,0 +1,79 @@
...
+
+use gtk::prelude::*;
+
+fn main() {
+    if gtk::init().is_err() {
+        println!("Failed to initialize GTK.");
+        return;
+    }
+
+    let window = gtk::Window::new(gtk::WindowType::Toplevel);
```
gkoz: Can we import `Window` and `WindowType` at the outset?

gkoz:
It's helpful to wrap the lines in the source to make line comments less ambiguous and small fixes more contained.

_tutorial/part_3.md
```diff
@@ -0,0 +1,39 @@
...
+// first we create a list with two columns
+GtkTreeModel *liststore = gtk_list_store_new(2, G_TYPE_STRING, G_TYPE_UINT);
+
+// then we cast it into a GtkTreeSortable object 
+GtkTreeSortable *sortable = GTK_TREE_SORTABLE(liststore);
+
+// implementation of sort functions...
+```
+
+You can't use such cast in Rust, so here is how you do it:
```
gkoz: More importantly you should not need it. If you need this cast in such circumstances it's a bug. In this example `ListStore::new` returns `ListStore`, which has `TreeSortableExt` implemented for it.

_tutorial/part_3.md
```diff
@@ -0,0 +1,39 @@
...
+
+```Rust
+// first we create a list with two columns
+let list_store = gtk::ListStore::new(&[Type::String, Type::U32]);
+
+// then we get the TreeSortable object
+let sortable = list_store.downcast::<TreeSortable>().unwrap();
```
gkoz: `TreeSortable` is not `IsA<ListStore>`, it's the other way around so I don't think this would compile.

docs-src/tutorial/part_1.md
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    window.show_all();
+    gtk::main();
+}
+```
+
+I think that the code is easy enough to understand for anyone.
+The only problem you might have is the `connect_delete_event` and the closure we pass to it.
+The `connect_delete_event` method is a callback when the "close window" event is spanned
```
gkoz: This sentence needs some of https://github.com/gtk-rs/examples/pull/76#issuecomment-195803264 treatment and overall reconsideration. E.g. the method is not a callback, it connects the closure to the signal. Does a "close window" event exist?

docs-src/tutorial/part_1.md
```diff
@@ -0,0 +1,103 @@
...
+    window.set_title("gtk-rs tutorial");
+    window.set_default_size(350, 70);
+
+    window.connect_delete_event(|_, _| {
+        gtk::main_quit();
+        Inhibit(false)
+    });
+
+    window.show_all();
+    gtk::main();
```
gkoz: This one is not easy enough for anyone,

docs-src/tutorial/part_1.md
```diff
@@ -0,0 +1,103 @@
...
+
+    window.set_title("gtk-rs tutorial");
+    window.set_default_size(350, 70);
+
+    window.connect_delete_event(|_, _| {
+        gtk::main_quit();
+        Inhibit(false)
+    });
+
+    window.show_all();
```
gkoz: It's important to note that widgets are invisible by default and forgetting to show them is a common mistake.

docs-src/tutorial/part_1.md
```diff
@@ -0,0 +1,103 @@
...
+
+```Rust
+|_, _| {
+    gtk::main_quit();
+    Inhibit(false)
+}
+```
+
+This closure takes two arguments but we ignore them since we don't use them.
+The `gtk::main_quit()` function indicates to *gtk* we want to end our application.
```
gkoz: More specifically, it stops the event loop, which should been mentioned in relation to `gtk::main()`

docs-src/tutorial/part_1.md
```diff
@@ -0,0 +1,103 @@
...
+And finally, the `Inhibit(false)` returned value indicates to *gtk* that this event must be
```
gkoz: that the default signal handler should be run after ours

gkoz: Also might make sense to mention that the default signal handler destroys the window.

gkoz:
Shouldn't it be part 2 that explains WTF signal handlers have `Fn + 'static` bounds and how do deal with that? :wink:

docs-src/tutorial/part_3.md
```diff
@@ -0,0 +1,45 @@
...
+GtkTreeModel *liststore = gtk_list_store_new(2, G_TYPE_STRING, G_TYPE_UINT);
+
+// then we cast it into a GtkTreeSortable object 
+GtkTreeSortable *sortable = GTK_TREE_SORTABLE(liststore);
+
+// implementation of sort functions...
+```
+
+You can't use such cast in Rust and shouldn't need to. If you need it in such circumstances,
+it's a bug. In the following example, we could just use `TreeSortable` method because
```
gkoz: So.. my comment was a message to you not a suggestion of the content. I don't expect the reader to understand "such circumstances". The "first principles" thing people need to know about GTK class hierarchy in Rust is that if `B` is a subclass of `A` and implements interface `I`, then `B` implements `AExt` and `IExt`; additionally if `B` has subclasses, it implements `BExt` and this is where one should look for `B`'s methods.

Contrasting Rust with C can just conclude that thanks to the above one doesn't need casts all over the place.

To explain casts you need plausible scenarios when they are actually necessary.

docs-src/tutorial/part_4.md
```diff
@@ -0,0 +1,38 @@
...
+###Clone
+
+The `Clone` trait from Rust has been implemented on every *gtk-rs*'s widgets. So simply:
+
+```Rust
+let label = gtk::Label::new("label");
+let label_ref = label.clone();
+```
+
+Obviously, `label` won't be duplicated. Internally, *gtk-rs* calls the [g_object_ref](https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html#g-object-ref) function. To make it simple, it allows to avoid an object to get destroyed if it's still used somewhere in your rust code. Once an object is destroyed, the `Drop` trait will call the [g_object_unref](https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html#g-object-unref) function which will decrement the number of references of the object, and remove it if there is no more references.
```
gkoz: You explain reference counting yet don't even mention it? How many different things are called "object" here? What is "remove"?

docs-src/tutorial/part_4.md
```diff
@@ -0,0 +1,38 @@
...
+
+Now let's talk about the last particularity of *gtk-rs*: the glib's `Value` type!
+
+###glib and Value
+
+The `Value` type is used to wrap constant data from any type to be displayed into some views (like a tree view for example). The big plus of this type is the safety it brings: you cannot get/set a value if it doesn't correspond to the type that you entered when you passed it done to the model/view.
```
gkoz: It's a dynamically-typed generic value. Any type that implements `StaticType`. The second setnence talks about the feature of models not `Value` itself. And we might actually remove some asserts making sets fail silently instead if panicing.

docs-src/tutorial/part_3.md
GuillaumeGomez: I use them in my `process-viewer`, but I'm not sure it'd be a good example...

gkoz: I bet you can easily remove them.

GuillaumeGomez: I like them. :'(

Susurrus:
I'll go through this and give me thoughts too now that I'm a little more familiar with this, but I wanted to ask, and it should be made clear in the introduction here as well, who the target audience is? Is this for people new to Rust, new to GTK, comfortable Rust devs trying to get into GUI programming, or coming from GTK (C, C++, Python, Vala, etc.) to gtk-rs specifically. Without deciding on who the target market is, it's going to be hard for us to agree on everything after that? If no one has thought much about this, IMHO I'd suggest targeting comfortable Rust devs trying to get into GTK programming who have a C background so they understand C concepts and documentation. This is a narrow niche of people who I would consider most likely to consider `gtk-rs`.

GuillaumeGomez:
That was the main points that made us not merge this. With a few steps back, I now think that we should make a specific introduction to every type of users:

- new to rust
- new to gtk
- new to rust and gtk

And then we would make a few common chapters.

Susurrus:
What about choosing a single type of user and targeting them first? Then as we add additional users we're targeting we can reuse common chapters. Given the small team working on this, I think it's too much effort to start out that modular from the beginning, IMHO.

GuillaumeGomez:
The thing to keep in mind is that we have time. We're not in a hurry. `gtk-rs` is now quite stable so we won't have the issues we had back then and finally, there isn't much to present.

Susurrus:
Fair enough. So what're the next steps then?

GuillaumeGomez:
Determining what each introduction should talk about. So here's what I have in mind:

### New to rust

Talking about cargo, creating the gtk-rs project, adding dependencies and the minimum needed.

### New to GTK

Explaining a bit how gtk works, its signal and inheritance systems.

### New to both

I think in here we can just link to the two previous chapters.

## Then

After that, we need to actually show how to create a basic application. How to handle signal (and more specifically helping users to use newtypes like `Rc` and `RefCell`.

Then we need to talk about the glade tool to generate UIs and how to use with gtk-rs.

## Conclusion

As you can see, it's *not much* on itself, but it needs a bit of time of reviewing to have something as easy and accessible as possible.

PR #15, opened 2016-03-24
Fix using features

EPashkin:
@gkoz  ?

### DIFF
--- a/index.md
+++ b/index.md
@@ -38,9 +38,9 @@ Include `gtk` in your `Cargo.toml` and set the minimal GTK version required by y
 {% assign gtk = site.data.crates | where: "name", "gtk" %}
 
 ~~~toml
-[dependencies]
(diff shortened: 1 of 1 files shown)


EPashkin:
Not tested that script part still works.

EPashkin:
Error before this PR `version components must be numeric` so old variant just tried load crate `features` from `crates.io` and not worked anyway.

gkoz:
Thanks for noticing this! Unfortunately there seems to be an issue with highlighting this kind of toml syntax (the highlighter thinks it's broken, at least on my machine), so we need to either choose the more usual one
```toml
[dependencies.gtk]
version = "..."
features = ["..."]
```
or remove the "toml" hint.

EPashkin:
It come from gitter

EPashkin:
Updated

gkoz:
Thanks!

PR #24, opened 2017-02-08
Link to Garta project

zaari:
Garta application relies on gtk-rs binding. It can be included in the list of projects using gtk.


GuillaumeGomez:
Wo! Looks really nice!

PR #28, opened 2017-03-04
Add post for new release

GuillaumeGomez:
cc @EPashkin

### DIFF
--- a/_data/crates.json
+++ b/_data/crates.json
@@ -1,30 +1,30 @@
 [
   {
     "name": "cairo-rs",
-    "max_version": "0.1.1"
+    "max_version": "0.1.2"
   },
   {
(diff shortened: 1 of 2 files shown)


EPashkin:
Update version in `_data/crates.json` (changes badges and first page using example)
Important: Added not only child properties but "almost all" normal properties too.

GuillaumeGomez:
Better? :)

EPashkin:
"all" properties not right, but may be it normal :wink:

GuillaumeGomez:
Oh you're right. I removed "all". Anything else?

EPashkin:
All seems ok

GuillaumeGomez:
Thanks for your review!

PR #30, opened 2017-03-11
Add usage latest version

EPashkin:
Found minor problem: code markdown for toml don't support sub-arrays (see temporary version in https://epashkin.github.io/)

In github this example looks fine:

## Using latest, not published version

Include `gtk` in your `Cargo.toml` and add `replace` ([see more about it](http://doc.crates.io/specifying-dependencies.html#overriding-dependencies)):

~~~toml
[dependencies.gtk]
version = "0.1.2"
features = ["v3_10"]

[replace]
"gtk:0.1.2" = { git = 'https://github.com/gtk-rs/gtk' }
~~~


GuillaumeGomez:
Thanks!

PR #36, opened 2017-06-11
Add sourceview

GuillaumeGomez:


index.md
```diff
@@ -13,6 +13,7 @@ layout: wide
 [GLib]: https://developer.gnome.org/glib/stable/
 [GTK]: https://developer.gnome.org/gtk3/stable/
 [Cairo]: http://cairographics.org/documentation/
+[GtkSourceView]: https://wiki.gnome.org/Projects/GtkSourceView
```
EPashkin: This won't do anything without using it :(

GuillaumeGomez: What do you mean?

EPashkin: As I understand this just footnote. Others used in "bindings for"

GuillaumeGomez: Ah right, well, I didn't put the doc for the sourceview yet. I'm going to do it and add sourceview into the documented crates at the same time.

EPashkin:
By way about docs:
It normal that cairo docs don't contains `create_from_png` ?
Or doc generator don't detect feature-gated modules?

GuillaumeGomez:
I think it doesn't. In the future we'll need to generate docs more carefully in order to have every features for all crates. (yep, it's quite annoying)

PR #42, opened 2017-08-03
add cross compile tutorial

etrombly:
Added a tutorial for cross compiling from Linux to Windows

### DIFF
--- a/tuto/cross.md
+++ b/tuto/cross.md
@@ -1,4 +1,8 @@
-The package installation steps are for Arch Linux, but should be similar for other distros.
+---
+layout: default
+---
+
+# Preparing rust
 
 First install rust normally with rustup. The next step will install the windows toolchain.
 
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Thanks so much about this!

@EPashkin: Seems ok for you? (I'll add the missing footer links once it's merged)

tuto/cross.md
```diff
@@ -0,0 +1,41 @@
+The package installation steps are for arch, but should be similar for other distros.
```
EPashkin: "Arch Linux" instead "arch" ?

tuto/cross.md
```diff
@@ -0,0 +1,41 @@
...
+
+Then set up the target in ~/.cargo/config
+
+    [target.x86_64-pc-windows-gnu]
+    linker = "x86_64-w64-mingw32-gcc"
+    ar = "x86_64-w64-mingw32-gcc-ar"
+
+Now install mingw and gtk libraries, this will take a while
+
+    pacaur -S mingw-w64-gcc mingw-w64-freetype2-bootstrap mingw-w64-cairo-bootstrap
```
EPashkin: Maybe change it to `pacman`? Or default package manager changed?

EPashkin:
@etrombly Thanks, now we will have documented answer to "How distribute gtk application?" 😉

tuto/cross.md
etrombly: mingw is in the AUR, so it won't work with pacman. I can try it in ubuntu in a bit so we can have instructions for both if you'd like.

EPashkin: Understand now.
Can you add comment about this, if AUR don't installed by default?

etrombly:
Updated the formatting a little, made the Arch section more clear. Also added instructions to add a program icon and change the theme to a windows 10 native look. If you want you should be able to adapt those sections pretty easily for someone building on windows, I haven't tried. Didn't end up adding Ubuntu instructions.

GuillaumeGomez:
Thanks a lot! (you even added footer urls, awesome!)

PR #48, opened 2017-10-17
Added Way Cooler org to projects list

Timidger:
Added Way Cooler, which primarily uses the Cairo and Pango bindings (and probably the other bindings in the future).

I added the org itself, since the extensions also use cairo / pango.


GuillaumeGomez:
Nice, thanks! Hope you enjoy using it! :)

PR #75, opened 2018-05-03
 The second example dose not compile for me. 

antoniocorbibellot:
Correcting a minor typo and adding WidgetExt to the 'use' line solves the compilation problems.


GuillaumeGomez:
Thanks!

PR #89, opened 2018-11-08
Enforce https on gtk-rs.org

GuillaumeGomez:
What do you think of it @sdroege and @EPashkin? Not very useful in itself but not we have a very "nice" **unsafe** in red in the url bar on chrome. So I assume it'd be better to enforce https to avoid it.


EPashkin:
Strange, but in firefox I already has redirect to https and just can't go to http://gtk-rs.org/ at all

sdroege:
Should just enable HSTS really instead of such a Javascript hack :)

GuillaumeGomez:
Ah nevermind. I successfully enforced https through github options. Thanks to both of you! :)

PR #98, opened 2019-02-21
[release] merging master-release-update into master

GuillaumeGomez:
cc @GuillaumeGomez @EPashkin @sdroege


sdroege:
For the release notes just a list of all PRs instead of something more useful for human readers? It would be more useful for human readers to have the important changes listed explicitly and explained what that means :)

GuillaumeGomez:
It just creates the PR without me doing anything. I update the blog post afterwards (I'm currently actually). Hold on a bit. :)

GuillaumeGomez:
Updated release blog post.

EPashkin:
Thanks, 👍

_posts/2019-02-21-new-release.md
```diff
@@ -0,0 +1,417 @@
...
+
+Hi everyone!
+
+It's time for a new release! Main adds/changes this time are:
+
+ * We added the generation of the `Atk` crate.
+ * We now generate functions taking callback as parameters.
+ * We improved the channels handling in `GLib`.
+ * The whole new `GString` type!
+ * The minimum Rust version supported is now the `1.31`.
```
sdroege: Should also mention here
* Updated minimum version of all libraries to the GNOME 3.14 versions, and API is available up to and including 3.30
* Subclassing support in GLib (link to example from the docs and in the examples repo, and mention that the glue in gio/gtk for this is still missing but exists for GStreamer already)

_posts/2019-02-21-new-release.md
```diff
@@ -0,0 +1,417 @@
...
+ * Even more bindings generated.
+
+Let's see those in details.
+
+#### Atk
+
+The Atk crate is about accessibility. We thought it was a miss not having it considering how important accessibility is, now it's fixed. You can see more information directly on the `Atk` repository: https://github.com/gtk-rs/atk
```
sdroege: `... how important accessibility is. [period here, not comma] Now it's fixed! You can **find** more`

sdroege: Also link to the example maybe?

_posts/2019-02-21-new-release.md
```diff
@@ -0,0 +1,417 @@
...
+Nothing fancy here: it takes the object we're running this function upon, a callback and some user data. Now here's the Rust version:
+
+```rust
+fn foreach<P: FnMut(&TreeModel, &TreePath, &TreeIter) -> bool>(
+    &self,
+    func: P,
+);
+```
+
+No more user data since closures can capture their environment, just the object and the callback (aka closure). Makes things a bit simpler and nicer to use.
```
sdroege: Maybe worth mentioning that we had many such functions before already? The actual new thing here is that we can generate bindings for these functions automatically now instead of having to do that manually :)

_posts/2019-02-21-new-release.md
```diff
@@ -0,0 +1,417 @@
...
+
+    // Returning false here would close the receiver
+    // and have senders fail
+    glib::Continue(true)
+});
+```
+
+#### `GString` type?
+
+This type has been created in order to prevent some useless copies to improve performances. For example, if you want to get a button label, you don't ownership over it, you just "get" it. Before the `GString` type, we'd clone the returned string and gave it back to the user. Now, we just hold a temporary reference over it, no more copy!
```
sdroege: The example is wrong. `Button::get_label()` is `transfer none`, so we copy here like before (and have to).

`GString` makes a difference if a string is returned `transfer full`. That means that C gives us a string copy, and previously we had to create yet another copy of that ourselves to produce a `String`. Now we simply wrap the copy we get from C in `GString` and don't do another copy.

_posts/2019-02-21-new-release.md
```diff
@@ -0,0 +1,417 @@
...
+
+This is part of our performance focus. More to come in the next release!
+
+#### Minimum Rust version supported
+
+We moved it to `1.31.0` mainly because imports handling is much easier starting this version.
```
sdroege: Worth mentioning https://github.com/gtk-rs/glib/issues/420 here and that macros (and the crates in general) still need to be moved to the 2018 edition.

_posts/2019-02-21-new-release.md
```diff
@@ -0,0 +1,417 @@
...
+
+#### Conclusion
+
+That's it for this release! Don't forget to take a look at our brand new [F.A.Q. section](https://gtk-rs.org/docs-src/faq).
+
+And again: thanks **a lot** to all of our contributors! This project lives thanks to their awesome work!
+
+### Changes
+
+For the interested ones, here is the list of the (major) changes:
```
sdroege: I'd say: "list of all the PRs that were merged" :) This is not the *major* changes, this is literally all the changes however small they are

_posts/2019-02-21-new-release.md
GuillaumeGomez: Good point!

_posts/2019-02-21-new-release.md
GuillaumeGomez: Ah crap, got confused again...

_posts/2019-02-21-new-release.md
GuillaumeGomez: Not all of them actually. I filter a few out. :)

GuillaumeGomez:
Updated.

PR #106, opened 2019-04-13
Update the GTK version number in the section Using

remontees:
Since version 0.6.0, the minimum version of GTK proposed is v3_16.


GuillaumeGomez:
Thanks!

PR #108, opened 2019-05-10
Gir tutorial

GuillaumeGomez:
I'll need https://github.com/gtk-rs/gir/pull/755 to be merged first.

cc @EPashkin @sdroege

### DIFF
--- a/docs-src/tutorial/gir_tutorial.md
+++ b/docs-src/tutorial/gir_tutorial.md
@@ -1,0 +1,558 @@
+---
+layout: default
+---
+
+# Generating a GNOME library using gir crate
+
+In this tutorial, we'll see how to generate a GNOME library using the [gir] crate. A few things to note first:
+
+ * It only works on GObject-based libraries.
+ * You need `.gir` files.
+
+A little explanation about those requirements: the [gir] crate needs `.gir` files to generate the library API. You can generally find them alongside the library header files (as you can see [here](https://packages.debian.org/stretch/amd64/libgtk-3-dev/filelist) for example, look for ".gir").
+
+The `.gir` files "describes" the library API (objects, arguments, even ownership!). This is where the [gir] crate comes in: it reads those `.gir` files and generates the Rust crates from them. You can learn more about the GIR format [here](https://gi.readthedocs.io/en/latest/).
+
+A little note about the `.gir` files: it often happens that they are invalid (missing or invalid annotations for example). We have a small script to fix the `.gir` files we're using (and only them!) available in the [gir-files repository](https://github.com/gtk-rs/gir-files/blob/master/fix.sh). You can run it like this (at the same level of the `.gir` files you want to patch):
+
+```bash
+> sh fix.sh
+```
+
+All `gtk-rs` generated crates come in two parts: the `sys` part which contains all the C functions and types definitions (direct mapping, everything is unsafe) and the "high-level" part which contains the nice, completely safe and idiomatic Rust API on top of the `sys` part.
+
+As an example, we'll generate the `sourceview` library bindings. So first, let's generate the `sys` part!
+
+## Generating the sys-library
+
+First, you'll need to download [gir]:
+
+```bash
+> git clone https://github.com/gtk-rs/gir
+> cd gir
+> cargo install --path . # so we can use gir binary directly
+```
+
+Then the `.gir` files (luckily for you, we have a repository which contains all the ones you need for sourceview!):
+
+```bash
+> git clone https://github.com/gtk-rs/gir-files
+```
+
+If you look into `gir-files`, you'll see a file named `GtkSource-3.0.gir`. That's the one for sourceview.
+
+Now let's create a new project for our `sourceview` crate:
+
+```bash
+> cargo new sourceview --lib
+```
+
+Then let's create a folder inside the newly created `sourceview` folder for the `sys` part:
+
+```bash
+> cd sourceview
+> cargo new sourceview-sys --lib
+```
+
+To indicate to [gir] what to generate, we'll need a `Gir.toml` file (inside the `sourceview-sys` folder) containing:
+
+```toml
+[options]
+library = "GtkSource"
+version = "3.0"
+target_path = "."
+min_cfg_version = "3.0"
+```
+
+ * `library` stands for the library we want to generate.
+ * `version` stands for the version of the library to be used.
+ * `target_path` stands for the location where the files will be generated.
+ * `min_cfg_version` will be the minimum version supported by the generated bindings.
+
+You should now have a folder looking like this:
+
+```text
+sourceview/
+  |
+  |---- Cargo.toml
+  |---- sourceview-sys/
+  |       |
+  |       |---- Cargo.toml
+  |       |---- Gir.toml
+  |       |---- src/
+  |               |
+  |               |---- lib.rs
+  |---- src/
+          |
+          |---- lib.rs
+```
+
+Let's generate the `sys` crate now:
+
+```bash
+> cd sourceview-sys
+# Run gir in "sys" mode (the "-m" option) and we give the gir files path (the "-d" option)
+> gir -m sys -d ../../gir-files/
+```
+
+(In case a failure happens at this point, and you can't figure out what's going on, don't hesitate to reach us so we can give you a hand!)
+
+You should now see new files (and a new folder):
+
+ * `build.rs`
+ * `Cargo.toml`
+ * `src/lib.rs`
+ * `tests/`
+
+Now let's try to build it:
+
+```bash
+> cargo build
+```
+
+Surprise! It doesn't build at all and you should see a loooooot of errors. Well, that was expected. We need to add some dependencies (you can find which ones in the `.gir` files) in order to make it work. Let's update our `Gir.toml` file to make it look like this:
+
+```toml
+[options]
+library = "GtkSource"
+version = "3.0"
+target_path = "."
+min_cfg_version = "3.0"
+
+external_libraries = [
+    "Cairo",
+    "Gdk",
+    "GdkPixbuf",
+    "Gio",
+    "GLib",
+    "GObject",
+    "Gtk",
+]
+```
+
+Now we regenerate it then rebuild it:
+
+```bash
+> rm Cargo.* # we remove Cargo files
+> gir -m sys -d ../../gir-files/
+> cargo build
+```
+
+Should work just fine!
+
+We can cleanup the command line a bit now. You can actually give the work mode ("-m" option) and the gir files repository through the `Gir.toml` file using "work_mode" and "girs_dir" options:
+
+```toml
+[options]
+library = "GtkSource"
+version = "3.0"
+target_path = "."
+min_cfg_version = "3.0"
+work_mode = "sys"
+girs_dir = "../../gir-files/"
+
+external_libraries = [
+    "Cairo",
+    "Gdk",
+    "GdkPixbuf",
+    "Gio",
+    "GLib",
+    "GObject",
+    "Gtk",
+]
+```
+
+Now, if you want to regenerate, just run:
+
+```bash
+> gir
+```
+
+Now we have a working `sys` containing all functions and objects definition. Just to be sure everything was correctly generated, we can run some tests (graciously generated by [gir] as well):
+
+```bash
+> cargo test
+```
+
+Normally, all tests passed. If you get an error when running those tests, it's very likely that the `sys` generation is invalid and/or incomplete.
+
+Time to generate the high-level Rust API!
+
+## Generating the high-level Rust API
+
+Time to go back to the "global" sourceview folder:
+
+```bash
+> cd ..
+```
+
+As you certainly guessed, we'll need a new `Gir.toml` file. Let's write it:
+
+```toml
+[options]
+girs_dir = "../gir-files"
+library = "GtkSource"
+version = "3.0"
+min_cfg_version = "3.0"
+target_path = "."
+work_mode = "normal"
+generate_safety_asserts = true
+deprecate_by_min_version = true
+single_version_file = true
+
+generate = []
+```
+
+A few new things in here. Let's take a look at them:
+
+ * "work_mode" value is now set to "normal", it means it'll generate the high-level Rust api instead of the sys-level.
+ * "generate_safety_asserts" is used to generates checks to ensure that, or any other kind of initialization needed before being able to use the library.
+ * "deprecate_by_min_version" is used to generate a [Rust "#[deprecated]"](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html) attribute based on the deprecation information provided by the `.gir` file.
+ * "single_version_file" is a very useful option when you have a lot of generated files (like we'll have). Instead of generating the gir hash commit used for the generation in the header of all generated files, it'll just write it inside one file, removing `git diff` noise **a lot**.
+ * "generate = []": this line currently does nothing. We say to [gir] to generate nothing. We'll fulfill it later on.
+
+Let's make a first generation of our high-level Rust API!
+
+```bash
+> gir
+```
+
+Now if you take a look around, you'll see a new "auto" folder inside "src". Doesn't contain much though. Which makes sense since we're generating nothing. Time to introduce you to a whole new [gir] mode: not_bound. Let's give it a try:
+
+```bash
+> gir -m not_bound
+[NOT GENERATED] GtkSource.Buffer
+[NOT GENERATED PARENT] Gtk.TextBuffer
+[NOT GENERATED] GtkSource.Language
+[NOT GENERATED] GtkSource.Mark
+[NOT GENERATED PARENT] Gtk.TextMark
+[NOT GENERATED] GtkSource.StyleScheme
+[NOT GENERATED] GtkSource.UndoManager
+[NOT GENERATED] GtkSource.SortFlags
+[NOT GENERATED] GtkSource.Completion
+[NOT GENERATED PARENT] Gtk.Buildable
+[NOT GENERATED] GtkSource.CompletionProvider
+[NOT GENERATED] GtkSource.CompletionContext
+[NOT GENERATED PARENT] GObject.InitiallyUnowned
+[NOT GENERATED] GtkSource.CompletionInfo
+[NOT GENERATED PARENT] Gtk.Window
+[NOT GENERATED PARENT] Gtk.Bin
+[NOT GENERATED PARENT] Gtk.Container
+[NOT GENERATED PARENT] Gtk.Widget
+[NOT GENERATED PARENT] Atk.ImplementorIface
+[NOT GENERATED] GtkSource.View
+[NOT GENERATED PARENT] Gtk.TextView
+[NOT GENERATED PARENT] Gtk.Scrollable
+[NOT GENERATED] GtkSource.CompletionActivation
+[NOT GENERATED] GtkSource.CompletionProposal
+[NOT GENERATED] GtkSource.CompletionError
+[NOT GENERATED] GtkSource.CompletionItem
+[NOT GENERATED PARENT] GtkSource.CompletionProposal
+[NOT GENERATED] GtkSource.CompletionWords
+[NOT GENERATED PARENT] GtkSource.CompletionProvider
+[NOT GENERATED] GtkSource.DrawSpacesFlags (deprecated in 3.24)
+[NOT GENERATED] GtkSource.Encoding
+[NOT GENERATED] GtkSource.File
+[NOT GENERATED] GtkSource.MountOperationFactory
+[NOT GENERATED] GtkSource.FileLoader
+[NOT GENERATED] GtkSource.FileLoaderError
+[NOT GENERATED] GtkSource.FileSaver
+[NOT GENERATED] GtkSource.FileSaverFlags
+[NOT GENERATED] GtkSource.FileSaverError
+[NOT GENERATED] GtkSource.Gutter
+[NOT GENERATED] GtkSource.GutterRenderer
+[NOT GENERATED] GtkSource.GutterRendererState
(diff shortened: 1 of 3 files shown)


docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+> cd sourceview-sys
+# Run gir in "sys" mode (the "-m" option) and we give the gir files path (the "-d" option)
+> gir -m sys -d ../../gir-files/
+```
+
+You should now see new files:
+
+ * `build.rs`
+ * `Cargo.toml`
+ * `src/lib.rs`
```
EPashkin: tests/* also generated

EPashkin: and versions.txt one or more depends config

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+
+To indicate to [gir] what to generate, we'll need a `Gir.toml` file (inside the `sourceview-sys` folder) containing:
+
+```toml
+[options]
+library = "GtkSource"
+version = "3.0"
+target_path = "."
+min_cfg_version = "3.0"
+```
```
EPashkin: IMHO better add `work_mode = "sys"`
and maybe `single_version_file = "."`

EPashkin: Ignore this, I don't read next part 😭

EPashkin:
@GuillaumeGomez Thanks, this tutorial is very good.

GuillaumeGomez:
@EPashkin Thanks for you feedbacks! I'll see the error you spotted!

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
+---
+layout: default
+---
+
+# Generating a GNOME library using gir crate
+
+In here, we'll see how to generate a GNOME library using the [gir] crate. A few things to note first:
```
gdesmott: "In this tutorial" maybe?

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+
+ * It only works on GNOME libraries.
```
gdesmott: "In GLib based libraries" ?

nielsdg: Could you possible define this a little bit more in-depth? :-)

For example, doest that mean the library won't run under windows? Or only that it has a GObject/GLib dependency?

ebassi: "It only works on GObject-based libraries"; being part of GNOME has nothing to do with that. :wink:

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+ * You need `.gir` files.
+
+A little explanation about those requirements: the [gir] crate needs `.gir` files to generate the library API. You can generally find them alongside the library header files (as you can see [here](https://packages.debian.org/stretch/amd64/libgtk-3-dev/filelist) for example, look for ".gir").
+
+The `.gir` files "describes" the library API (objects, arguments, even ownership!). This is where the [gir] crate comes in: it reads those `.gir` files and generates the Rust crates from them.
+
```
gdesmott: Would be good point to some gir documentation for those not use to it. https://gi.readthedocs.io/en/latest/ maybe?

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: No problem. ;)

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+library = "GtkSource"
+version = "3.0"
+target_path = "."
+min_cfg_version = "3.0"
+```
+
+ * `library` stands for the library we want to generate.
+ * `version` stands for the version of the library to be used.
+ * `target_path` stands for the location where the files will be generated.
+ * `min_cfg_version` is the minimum "feature" that will be generated.
```
gdesmott: This is not that clear to me. What are those features?

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Aren't `.gir` files generated only for GNOME libraries? How would you word it?

GuillaumeGomez: I asked about it [here](https://github.com/gtk-rs/gtk-rs.github.io/pull/108#discussion_r282880812). It apparently needs more precisions indeed.

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Indeed! Applying change.

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+generate = []
+```
+
+A few new things in here. Let's take a look at them:
+
+ * "work_mode" value is now set to "normal", it means it'll generate the high-level Rust api instead of the sys-level.
+ * "generate_safety_asserts" is used to generates checks to ensure that the GTK initialization functions have been called before the current one(s).
+ * "single_version_file" is a very useful option when you have a lot of generated files (like we'll have). Instead of generating the gir hash commit used for the generation in the header of all generated files, it'll just write it inside one file, removing `git diff` noise **a lot**.
+ * "generate = []": this line currently does nothing. We say to [gir] to generate nothing. We'll fulfill it later on.
+
```
gdesmott: You didn't document `deprecate_by_min_version`

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+```bash
+> gir
+```
+
+We now have a `src/auto/language.rs` file. We need to include all `auto` files in our library. To do so, let's update the `src/lib.rs` file as follows:
+
+```rust
+pub use auto::*;
+
+mod auto;
```
gdesmott: Shouldn't it be the other way around? `mod` first and then `pub use`?

docs-src/tutorial/gir_tutorial.md
ebassi: GNOME libraries (typically) generate introspection data, but any GObject-based library can (and should) do so, regardless of project affiliation. GObject-based libraries are not an exclusive of GNOME; XFCE has them, Elementary has them, and there are a bunch of low level libraries hosted on FreeDesktop.org that use GObject.

I'd say exactly that: it only works for GObject-based libraries that generate introspection data. If you want, you could like "GObject" to an online resource, like https://wiki.gnome.org/Projects/GLib.

GuillaumeGomez: I'll do it then. Thanks!

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+
+There are a few kinds of errors (not much luckily) which can happen with [gir] generation. Let's take a look at them.
+
+### Missing memory management functions
+
+If [gir] generation fails (for whatever reason), it means you'll have to implement the type yourself. Just like types from other `gtk-rs` crates, you'll need to put it into the "manual" list. Then you need to put the type into the `src` folder (or inside a subfolder, you know how Rust works).
+
+/!\ Don't forget to reexport the type inside your `src/lib.rs` file! For example, let's take a look at the [requisition.rs](https://github.com/gtk-rs/gtk/blob/master/src/requisition.rs) file from the `gtk` crate.
+
+Since it's a "simple" type (no pointer, therefore no memory management to do), [gir] doesn't know how to generate it. You'll need to implement some traits by hand like `ToGlibPtr` or `ToGlibPtrMut` (depending on your needs).
```
gdesmott: That's actually much harder than it sounds but that get explained in another tutorial I guess. :)

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Good point!

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Good catch!

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Nope, works both way.

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Well, no. But that would deserve a tutorial on its own. :-.

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,529 @@
...
+girs_dir = "../gir-files"
+library = "GtkSource"
+version = "3.0"
+min_cfg_version = "3.0"
+target_path = "."
+work_mode = "normal"
+generate_safety_asserts = true
+deprecate_by_min_version = true
+single_version_file = true
+
```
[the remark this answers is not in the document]
GuillaumeGomez: It's documented in the gir readme.

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
+---
+layout: default
+---
+
+# Generating a GNOME library using gir crate
+
+In this tutorial, we'll see how to generate a GNOME library using the [gir] crate. A few things to note first:
+
+ * It only works on GObject-based libraries.
+ * You need `.gir` files.
```
sdroege: Maybe a link to https://gi.readthedocs.io/en/latest/ ?

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+
+A little explanation about those requirements: the [gir] crate needs `.gir` files to generate the library API. You can generally find them alongside the library header files (as you can see [here](https://packages.debian.org/stretch/amd64/libgtk-3-dev/filelist) for example, look for ".gir").
+
+The `.gir` files "describes" the library API (objects, arguments, even ownership!). This is where the [gir] crate comes in: it reads those `.gir` files and generates the Rust crates from them. You can learn more about the GIR format [here](https://gi.readthedocs.io/en/latest/).
```
sdroege: It might be worth mentioning somewhere that fixes to the .gir generation, to the C code (annotations), etc. might be needed. Or fixing the .gir XML file with the xpath fixup thingy we use in the gir-files repo.

So that people are not directly demotivated that `gir` explodes on their .gir file because it's malformed... as very often happens unfortunately.

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+
+All `gtk-rs` generated crates come in two parts: the `sys` part which contains all the C functions and types definitions and the "high-level" part which contains the nice Rust API.
```
sdroege: Maybe mention that the sys part provides a direct mapping of the C API, only FFI functions, everything unsafe. And that the non-sys part contains completely safe, idiomatic Rust on top of the sys bindings? Make it more explicit what the difference between the two is

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+library = "GtkSource"
+version = "3.0"
+target_path = "."
+min_cfg_version = "3.0"
+```
+
+ * `library` stands for the library we want to generate.
+ * `version` stands for the version of the library to be used.
+ * `target_path` stands for the location where the files will be generated.
+ * `min_cfg_version` is the minimum version "feature" that will be generated (`gtk-rs` handles different libraries versions through `cargo` features).
```
sdroege: + "This will be the minimum version supported by the generated bindings."

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+          |
+          |---- lib.rs
+```
+
+Let's generate the `sys` crate now:
+
+```bash
+> cd sourceview-sys
+# Run gir in "sys" mode (the "-m" option) and we give the gir files path (the "-d" option)
+> gir -m sys -d ../../gir-files/
```
sdroege: At this point, gir might panic or give errors. Mention that and what the user should do then :) (Talk to us!)

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+ * `src/lib.rs`
+ * `tests/`
+
+Now let's try to build it:
+
+```bash
+> cargo build
+```
+
+Surprise! It doesn't build at all and you should see a loooooot of errors. Well, that was expected. We need to add some dependencies in order to make it work. Let's update our `Gir.toml` file to make it look like this:
```
sdroege: Where would the user get the list of dependencies from? Hint: they're written in the .gir file :)

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+]
+```
+
+Now, if you want to regenerate, just run:
+
+```bash
+> gir
+```
+
+Now we have a working `sys` containing all functions and objects definition. Time to generate the high-level Rust API!
```
sdroege: There are also tests generated and might be worth running them to check if things are generated correctly :)

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+deprecate_by_min_version = true
+single_version_file = true
+
+generate = []
+```
+
+A few new things in here. Let's take a look at them:
+
+ * "work_mode" value is now set to "normal", it means it'll generate the high-level Rust api instead of the sys-level.
+ * "generate_safety_asserts" is used to generates checks to ensure that the GTK initialization functions have been called before the current one(s).
```
sdroege: The GTK initialization, or any other kind of initialization needed before being able to use the library.

(For GStreamer for example this has nothing to do with GTK but instead `gst_init()`)

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+ * "deprecate_by_min_version" is used to generate a [Rust "#[deprecated]"](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html) attribute.
```
sdroege: ... based on the deprecation information provided by the .gir file.

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+[NOT GENERATED PARENT] GtkSource.StyleSchemeChooser
+[NOT GENERATED] GtkSource.StyleSchemeChooserInterface
+[NOT GENERATED] GtkSource.StyleSchemeChooserWidget
+[NOT GENERATED] GtkSource.StyleSchemeManager
+[NOT GENERATED] GtkSource.Tag
+[NOT GENERATED PARENT] Gtk.TextTag
+[NOT GENERATED] GtkSource.ViewGutterPosition
+```
+
+We now have the list of all the non-yet generated items. Quite convenient!
```
sdroege: What does the `PARENT` mean? Explain :)

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+
+Let's start by generating one type. Let's update the "generate" array as follows:
```
sdroege: How do we select the first type we want to generate? It's useful to know how the library works, you ideally select the most independent and basic type. Something that does not depend on any other things that you didn't generate yet as far as possible.

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+
+    match fn {
+        get_type => || gtk_source_sys::gtk_source_language_get_type(),
+    }
+}
+```
+
+This macro comes from the `glib` crate. We didn't import it, therefore the Rust compiler can't find it. We'll also need its `sys` part (the case of `glib` is a bit special).
+
+Another issue that will arise is that we didn't import the `sourceview-sys` crate either. Alongside those two (three if we count `glib-sys`!), we'll need both `libc` and `bitflags`. Let's fix all of those issues at once! For that, we need to update the `Cargo.toml`:
```
sdroege: We don't need libc and bitflags *yet* though

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+
+[dependencies.glib-sys]
+git = "https://github.com/gtk-rs/sys" # all gtk-rs sys crates are in the sys repository
+```
+
+And to import those crates into `src/lib.rs`:
+
+```rust
+#[macro_use]
+extern crate glib;
```
sdroege: Do we want to do this Rust 2018 edition style already? The macro imports and all that are 2015 edition currently

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+mod auto;
+```
+
+Let's try to rebuild:
+
+```bash
+> cargo build
+```
+
+It worked! We have generated the `Language` item! I'll let you take a look at the `src/auto/language.rs` file for the interested ones, then we can continue.
```
sdroege: Mention that there can be commented out things and why that could happen and what to do then

sdroege: Also this could still fail to compile for a different library... if gir generates broken code or we simply don't support this case yet. And then people should talk to us :)

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+```toml
+generate = [
+    "GtkSource.Language",
+]
+```
+
+Another `gir` run:
+
+```bash
+> gir
```
sdroege: Mention that for some other library this could fail for various reasons. Broken gir file among them. And people should talk to us then instead of giving up :)

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+Everything is working, yeay! If you take another look at `src/auto/region.rs`, you'll see a lot less commented functions. Amongst the remaining ones, you'll see this one:
+
+```rust
+//#[cfg(any(feature = "v3_22", feature = "dox"))]
+//fn get_start_region_iter(&self, iter: /*Ignored*/RegionIter);
+```
+
+If a type name isn't prepend by `[crate_name]::`, then it means it comes from the current crate. To add it, just put it into the "generate" list of `Gir.toml`.
+
+At this point, you should have almost everything you need. There is just one last case we need to talk about.
```
sdroege: Mention maybe examples for ignoring functions and how to manually implement them. That's a very common case

sdroege: Oh you do that below, nevermind

docs-src/tutorial/gir_tutorial.md
```diff
@@ -0,0 +1,533 @@
...
+    pub fn is_empty(&self) -> bool {
+        // blablabla
+        true
+    }
+}
+```
+
+You might wonder: "why not just implementing it on the `Region` type directly?". Because like this, a subclass will also be able to use this trait easily as long as it implements `IsA<Region>`. For instance, in gtk, everything that implements `IsA<Widget>` (so almost every GTK types) can use those methods.
+
+As usual, don't forget to reexport the trait. A little tip about reexporting manual traits: in `gtk-rs`, we create a `src/prelude.rs` file which reexports all manual traits, making it simpler for users to use them. It looks like this:
```
sdroege: And mention that the prelude is only for re-exporting traits and meant in user code to be wildcard imported via `use prelude::*`

sdroege:
Looks great generally, thanks for working on this :) And feel free to ignore any comments of mine if you disagree, it's all minor stuff

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Yes sir!

GuillaumeGomez: Hum actually, I'd rather not. The gir errors and how to solve them would be an even longer tutorial than this one...

sdroege: I mean to mention that errors can happen, and if they do they shouldn't give up but rather talk to us :)

GuillaumeGomez: Oh I see. Good point!

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Ok!

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Yes sir!

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: Yes and no. In my case, I just pick one completely randomly and here we go. If you want to "enforce" it, I can add it, but this isn't relevant enough from my point of view.

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: True but we'll need them anyway sooooo. :p

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: We can update the tutorial later on. For now, we didn't made the upgrade so it wouldn't make sense.

docs-src/tutorial/gir_tutorial.md
GuillaumeGomez: I come back on the commented things later on. But I'll say something about broken generation here (again :p).

GuillaumeGomez:
Thanks everyone for your feedbacks and reviews!

PR #110, opened 2019-06-11
Update the front-page example

ketsuban:
The front page still shows an example which creates a top-level window and calls GTK functions like `gtk::init` and `gtk::main` directly, which isn't recommended.

In addition, the section on getting `gtk-rs` from Git rather than crates.io has a strangely old GTK version feature, so I changed it to `v3_16` to match the rest of the site.

### DIFF
--- a/index.md
+++ b/index.md
@@ -56,34 +56,30 @@ extern crate gtk;
 use gtk::prelude::*;
 ~~~
 
-Create a window, etc.
+Create an application, etc.
 
 ~~~rust
-use gtk::{Button, Window, WindowType};
+use gtk::{Application, ApplicationWindow, Button};
 
 fn main() {
-    if gtk::init().is_err() {
-        println!("Failed to initialize GTK.");
-        return;
-    }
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
This example is a bit outdated, indeed. Thanks!

cc @EPashkin @sdroege

index.md
```diff
@@ -94,7 +90,7 @@ Include `gtk` in your `Cargo.toml` not as crate but from git:
 ~~~toml
 [dependencies.gtk]
 git = "https://github.com/gtk-rs/gtk"
-features = ["v3_10"]
+features = ["v3_16"]
```
sdroege: This shouldn't be needed at all. The example should run without setting any features.

ketsuban: It's not actually related to the example. I just thought it was weird that this one place on the site has `v3_10` when every other instance of setting `features` uses `v3_16`.

sdroege: Ok :)

EPashkin:
@ketsuban Thanks, 👍

PR #111, opened 2019-06-12
Add instructions for installing on MSVC

ketsuban:
@sdroege suggested I add these instructions. I won't be in the least bit offended if you want to reformat them to match a style guide or anything like that.

### DIFF
--- a/docs-src/requirements.md
+++ b/docs-src/requirements.md
@@ -29,6 +29,45 @@ The __gtk__ crate expects __GTK+__, __GLib__ and __Cairo__ development files to 
 
 ## Windows
 
+On Windows Rust can use either the MSVC toolchain or the GNU toolchain. The latter is easier to
+install, but the former provides a better overall experience and allows interfacing with other
+Windows libraries.
+
+### MSVC toolchain
+
+1. Install GTK using [the instructions on the GTK website](https://www.gtk.org/download/windows.php).
+   This will take some time. Once it's finished, make a note of the folder `installed\x86-windows`
+   (32-bit)/`installed\x64-windows` (64-bit); these instructions will refer to it as `%VCPKGDIR%`.
+2. Add `%VCPKGDIR%\bin` to the beginning of your `PATH` environment variable.
+3. Set the `GTK_LIB_DIR` environment variable to `%VCPKGDIR%\lib`.
+4. Create the following symlinks. (NTFS has supported symlinks since Windows Vista, but prior to the
+   Windows 10 Creators Update it required elevated privileges. Copying the files also works, but
+   you'll need to do it again when GTK is updated.)
+    * `%VCPKGDIR%\lib\gtk-3.0.lib` points to `%VCPKGDIR%\lib\gtk-3.lib`.
+    * `%VCPKGDIR%\lib\gdk-3.0.lib` points to `%VCPKGDIR%\lib\gdk-3.lib`.
+    * `%VCPKGDIR%\bin\gtk-3.0.dll` points to `%VCPKGDIR%\bin\gtk-3.dll`.
+    * `%VCPKGDIR%\bin\gdk-3.0.dll` points to `%VCPKGDIR%\bin\gdk-3.dll`.
+5. Create the folder `%VCPKGDIR%\etc\gtk-3.0`. Inside it, create the file `settings.ini`:
+
+```ini
+[Settings]
+gtk-theme-name=win32
+```
+
+#### Possible problems
+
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Thanks!

docs-src/requirements.md
```diff
@@ -29,6 +29,45 @@ The __gtk__ crate expects __GTK+__, __GLib__ and __Cairo__ development files to
...
+   you'll need to do it again when GTK is updated.)
+    * `%VCPKGDIR%\lib\gtk-3.0.lib` points to `%VCPKGDIR%\lib\gtk-3.lib`.
+    * `%VCPKGDIR%\lib\gdk-3.0.lib` points to `%VCPKGDIR%\lib\gdk-3.lib`.
+    * `%VCPKGDIR%\bin\gtk-3.0.dll` points to `%VCPKGDIR%\bin\gtk-3.dll`.
+    * `%VCPKGDIR%\bin\gdk-3.0.dll` points to `%VCPKGDIR%\bin\gdk-3.dll`.
+5. Create the folder `%VCPKGDIR%\etc\gtk-3.0`. Inside it, create the file `settings.ini`:
+
+```ini
+[Settings]
+gtk-theme-name=win32
```
alatiera: Why is this needed? Adwaita isn't working?

ketsuban: [`vcpkg` doesn't install themes](https://github.com/microsoft/vcpkg/issues/6554). I couldn't find any official instructions for acquiring them manually, and I didn't think telling people to manually download MinGW packages on the official website was very professional.

alatiera: Adwaita has been part of GTK itself for a while now, it comes as is when building gtk nothing else needed. If its broken that's a serious bug that should be investigated, not worked around.

ketsuban: Be my guest. Heckling Microsoft over their packaging infrastructure is outside my remit.

alatiera: Oh wow, this is vcpkg, not just an msvc build. Yea, this was broken the moment it landed there, I am surprised it even runs. They did some weird autotools -> cmake translation without verifying a thing.

Apologies, thought this was linking against a local msvc build.

sdroege: >  If its broken that's a serious bug that should be investigated, not worked around.

vcpkg is known to be broken since a long time and that's why we didn't want to recommend using it. But as it's apparently still easier than the alternatives on Windows...

If someone wants to report it and list what exactly is needed: https://github.com/microsoft/vcpkg/issues/new

ketsuban: I maintain that it's [this issue](https://github.com/microsoft/vcpkg/issues/6554) and that [this](https://github.com/microsoft/vcpkg/issues/4417) is a duplicate, but I've had no useful feedback so I don't really have any idea how to get some eyeballs on the problem.

alatiera: I don't think we can fix their vcpkg issue in any sane way, they took the autotools build and bodged it to something that wouldn't even launch when they fist announced vcpkg.

It might be easier to get them to use the meson build that has since landed for gtk3 instead.

docs-src/requirements.md
```diff
@@ -29,6 +29,45 @@ The __gtk__ crate expects __GTK+__, __GLib__ and __Cairo__ development files to
...
+```
+
+#### Possible problems
+
+##### `unresolved external symbol gtk_font_chooser_level_get_type`
+
+This is [a bug](https://github.com/gtk-rs/gtk/issues/794) in version 0.6 of gtk-rs with versions of
+GTK prior to 3.22.30; either use gtk-rs from Git directly or backport [the fix](https://github.com/gtk-rs/gtk/pull/804)
+into the copy of gtk-rs in your Cargo registry.
```
sdroege: This can be removed now, 0.7.0 was released a while ago. Who makes a new PR? :)

docs-src/requirements.md
```diff
@@ -29,6 +29,45 @@ The __gtk__ crate expects __GTK+__, __GLib__ and __Cairo__ development files to
 
 ## Windows
 
+On Windows Rust can use either the MSVC toolchain or the GNU toolchain. The latter is easier to
+install, but the former provides a better overall experience and allows interfacing with other
+Windows libraries.
```
sdroege: This part is also wrong. You can interface with other Windows libraries just fine when using the GNU toolchain, as long as it's not a C++ library. But you can't interface with a C++ library directly from Rust anyway, so ...

PR #113, opened 2019-06-15
Fix syntax error and update example on homepage

wezm:
I tried building the code on the homepage and ran into some issues. This PR updates the example code so that it compiles and runs.

### DIFF
--- a/index.md
+++ b/index.md
@@ -37,23 +37,29 @@ layout: wide
 
 Prepare your system: [Requirements](http://gtk-rs.org/docs-src/requirements.html)
 
-Include `gtk` in your `Cargo.toml` and set the minimal GTK version required by your project:
+Include `gtk` and `gio` in your `Cargo.toml` and set the minimal GTK version required by your project:
 {% assign gtk = site.data.crates | where: "name", "gtk" %}
 
 ~~~toml
 [dependencies.gtk]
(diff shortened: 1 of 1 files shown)


index.md
```diff
@@ -62,12 +68,13 @@ Create an application, etc.
...
 fn main() {
-    let application = Application::new("com.github.gtk-rs.examples.basic", Default::default()))
+    let application = Application::new("com.github.gtk-rs.examples.basic", Default::default())
         .expect("failed to initialize GTK application");
 
     application.connect_activate(|app| {
-        let window = ApplicationWindow::new();
+        let window = ApplicationWindow::new(app);
         window.set_title("First GTK+ Program");
+        window.set_border_width(10);
```
GuillaumeGomez: I'm not sure this line is very useful in this example. ;°

wezm: Perhaps not but I was mostly just updating it to do what https://github.com/gtk-rs/examples/blob/master/src/bin/basic.rs does.

GuillaumeGomez: The goal is to remain as small as possible in here to just let users see a sample. Please remove this line and then I merge.

wezm: Fixed :)

GuillaumeGomez:
Thanks!

PR #119, opened 2019-07-08
Remove obsolete "known problem" from Windows instructions

sdroege:
Also clarify a bit that it's not interfacing with other Windows
libraries that is possible with the MSVC toolchain, but only with ones
that are incompatible with MINGW. Which are not that many after all.

And add a comment that using the MSVC toolchain allows using the Visual
Studio debugger.

### DIFF
--- a/docs-src/requirements.md
+++ b/docs-src/requirements.md
@@ -30,8 +30,9 @@ The __gtk__ crate expects __GTK+__, __GLib__ and __Cairo__ development files to 
 ## Windows
 
 On Windows Rust can use either the MSVC toolchain or the GNU toolchain. The latter is easier to
-install, but the former provides a better overall experience and allows interfacing with other
-Windows libraries.
+install, but the former provides a better overall experience and allows interfacing with Windows
+libraries that are incompatible with the GNU toolchain and allows for example usage of the
+Visual Studio debugger.
 
 ### MSVC toolchain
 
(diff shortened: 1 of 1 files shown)


docs-src/requirements.md
```diff
@@ -30,8 +30,9 @@ The __gtk__ crate expects __GTK+__, __GLib__ and __Cairo__ development files to
 ## Windows
 
 On Windows Rust can use either the MSVC toolchain or the GNU toolchain. The latter is easier to
-install, but the former provides a better overall experience and allows interfacing with other
-Windows libraries.
+install, but the former provides a better overall experience and allows interfacing with Windows
+libraries that are incompatible with the GNU toolchain and for example usage
```
GuillaumeGomez: "and for example usage of the..." sounds a bit weird to me?

sdroege: [...] and allows [thing one] and for example usage of [...]

Should I repeat the verb?

sdroege: Yes? :)

EPashkin:
@sdroege Thanks

docs-src/requirements.md
GuillaumeGomez: Yes! Sorry. ^^'

sdroege: Done

GuillaumeGomez:
Thanks!

PR #123, opened 2019-08-30
Remove unused channels

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/docs-src/contact.md
+++ b/docs-src/contact.md
@@ -8,14 +8,11 @@ There are multiple ways to contact directly `Gtk-rs` developers:
 
 # IRC
 
-In here you have the choice between two channels on two different servers:
+Channel #rust on **irc.gimp.org** server.
 
- * #gtk-rs on irc.mozilla.org
- * #rust on irc.gimp.org
+# Matrix
 
-# Gitter
-
-You can find us on gitter at this address: [gitter.im/gtk-rs/gtk](https://gitter.im/gtk-rs/gtk).
+You can connect using: **#gimpnet#gnome:matrix.org**.
 
 # Github
 


docs-src/contact.md
```diff
@@ -8,14 +8,7 @@ There are multiple ways to contact directly `Gtk-rs` developers:
...
 
-In here you have the choice between two channels on two different servers:
-
- * #gtk-rs on irc.mozilla.org
- * #rust on irc.gimp.org
-
-# Gitter
-
-You can find us on gitter at this address: [gitter.im/gtk-rs/gtk](https://gitter.im/gtk-rs/gtk).
+#rust on irc.gimp.org
```
sdroege: Why make this a headline? Shouldn't it just be normal test, and ideally with some more information how to get there and maybe also how to use it from Matrix? :)

GuillaumeGomez: Oh crap. I'll rephrase it a bit to avoid this heading issue. Also, I don't use Matrix so it'll need someone using it to add instructions. :)

EPashkin: @GuillaumeGomez why gitter was removed?

GuillaumeGomez: Is anyone still using gitter?

sdroege: I always forget to get online on gitter again. But there are people and they're sometimes asking questions :) Probably not so active because none of us is very active there.

For Matrix, you've already put instructions for that into the topic of the IRC channel on the Mozilla network. Just copy that!

GuillaumeGomez:
Updated.

sdroege:
:+1: Unless someone feels like gitter should stay (but then that person should also stay on gitter and help people there).

EPashkin:
👍

GuillaumeGomez:
Then I merge. Adios gitter! :)

PR #126, opened 2019-10-10
remove and update expired  links

mr-tron:


sdroege:
That was fixed as part of https://github.com/gtk-rs/gtk-rs.github.io/pull/133 and https://github.com/gtk-rs/gtk-rs.github.io/pull/134

PR #139, opened 2020-02-19
Point to gnome installation page instead of ours

GuillaumeGomez:
As discussed:

cc @sdroege


docs-src/requirements.md
```diff
@@ -1,160 +0,0 @@
...
- *  Start the "MSYS2 MinGW Shell" (not to be confused with "MSYS2 MSYS Shell").
-
-### Possible problems
-
-#### Error on linking on Windows MSYS2: undefined reference to `__imp___acrt_iob_func'
-
-Download these packages and install it with `pacman -U <path>`:
-
-* http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-crt-git-5.0.0.5002.34a7c1c0-1-any.pkg.tar.xz
-* http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-headers-git-5.0.0.5002.34a7c1c0-1-any.pkg.tar.xz
```
EPashkin: Maybe this solution need to be saved.

docs-src/index.md
```diff
@@ -3,7 +3,9 @@ layout: default
 ---
 # Gtk-rs documentation
 
-## [Requirements](requirements.html)
+## Requirements
+
+First, prepare your system by taking a look at the [GNOME installation page](https://www.gtk.org/docs/installations/).
```
sdroege:
```suggestion
First, prepare your system by taking a look at the [GTK installation page](https://www.gtk.org/docs/installations/).
```

Where did you get the `GNOME` from? :)

docs-src/requirements.md
GuillaumeGomez: Should be done on the gtk instructions directly maybe then?

GuillaumeGomez:
Updated.

docs-src/requirements.md
EPashkin: Maybe or in FAQ

GuillaumeGomez: Definitely on the gtk website. :p

EPashkin: Oh you mean't https://www.gtk.org.
No, it rust problem not gtk's, due old msys32 in rust, so it better be somewhere in your site

GuillaumeGomez: Oh I see! Huuuuuum... Then I don't really know where to put it to make it easy to find...

EPashkin: Seems this part actually don't needed anymore,
I updated mingw2_64 and examples now builds fine 🎉

EPashkin: Seems issue was fixed by https://github.com/rust-lang/rust/pull/67429 only 16 days ago 😉

GuillaumeGomez: Yeay!

PR #142, opened 2020-03-10
mention another gtk-rs application, cigale

emmanueltouzery:
otherwise cigale uses relm on top of gtk-rs.


GuillaumeGomez:
Thanks!

PR #148, opened 2020-04-06
Contact page: Update matrix room & add a link

jplatte:
The room id that's currently mentioned on the page doesn't actually exist. `#rust:gnome.org` is the canonical alias for the matrix room bridged to `#rust` on GIMPnet.

### DIFF
--- a/docs-src/contact.md
+++ b/docs-src/contact.md
@@ -12,7 +12,9 @@ Channel #rust on **irc.gimp.org** server.
 
 # Matrix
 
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
I'm not sure if anyone from gtk-rs is using matrix... But thanks in any case!

jplatte:
Well, since the room is bridged, it doesn't matter whether other people are using IRC or matrix :)

GuillaumeGomez:
Absolutely!

PR #151, opened 2020-04-12
Some updates

hfiguiere:
Two commits:
- Update the link to gpsami as it is now on Gnome gitlab
- Update the crate versions.

I could not test it as jekyll + Ruby decided to be very uncooperative.

### DIFF
--- a/_data/crates.json
+++ b/_data/crates.json
@@ -1,11 +1,11 @@
 [
   {
     "name": "cairo-rs",
-    "max_version": "0.8.0"
+    "max_version": "0.8.1"
   },
   {
     "name": "gdk",
-    "max_version": "0.12.0"
+    "max_version": "0.12.1"
   },
   {
     "name": "gdk-pixbuf",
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Thanks for the update! However, I don't see the point of adding all these crates to the list. Most of them aren't meant to be used by end-users, therefore it's definitely not worth adding them. About the GNOME v4 crates, since they're absolutely not stable for the moment, better not adding them. As for the non-gtk-rs crates, we definitely don't want them in the version list.

hfiguiere:
I ran the update script... so that's perobably where the problem lies.

hfiguiere:
The script being `update_crates.sh`.

GuillaumeGomez:
I have to admit I didn't know/remember about this script...

hfiguiere:
I have removed the added crates by hand. This script needs fixing in some way or be removed.

GuillaumeGomez:
Thanks!

PR #154, opened 2020-06-16
Document the object-oriented nature of gtk-rs

piegamesde:
The old "Upcast and downcast" article already touched that topic, so I decided to rename and extend that page.


docs-src/tutorial/object_oriented.md
```diff
@@ -2,13 +2,23 @@
...
+# Object-oriented Rust
 
-Since there is an inheritance system in Gtk, it's only logical to have one as well in `Gtk-rs`. Normally, most people won't need this, but it's always nice to understand how things work.
+Since there is an inheritance system in Gtk, it's only logical to have one as well in `Gtk-rs`. Normally, most people won't need this, but understanding how things work can help navigating the documentation a lot.
+
+Each GTK class is split into a struct named after that class, and a trait with the same name and the suffix `Ext`. All methods except the constructor(s) are not in the struct, but in that trait. **Notice 1:** some methods are hidden in a `…ExtManual` trait, as seen in [WidgetExtManual](http://gtk-rs.org/docs/gtk/prelude/trait.WidgetExtManual.html). **Notice 2:** The same principle applies to GTK interfaces as well (have a look at the [Orientable](http://gtk-rs.org/docs/gtk/struct.Orientable.html) interface for an example).
+
+## Basic inheritance
+
+Any struct not only implements it's `Ext` trait, but also all traits of its super classes. That way you can call methods of super classes directly on a struct.
```
GuillaumeGomez, line 13: "it's" -> "its"

docs-src/tutorial/object_oriented.md
```diff
@@ -2,13 +2,23 @@
...
+
+The [`IsA`](http://gtk-rs.org/docs/glib/object/trait.IsA.html) trait is used to model the subtype relation of classes between structs. Every struct implements `IsA<SuperClass>` for each of its parent classes (and its implemented interfaces). For example [FlowBox](http://gtk-rs.org/docs/gtk/struct.FlowBox.html) implements `IsA<Buildable>`, `IsA<Container>`, `IsA<Orientable>`, `IsA<Widget>`.
+
+Passing classes as arguments it pretty straightforward. The only rule is that when accepting a class as function argument, take a generic `IsA<ClassThatIWant>` instead: [`fn add<P: IsA<Widget>>(&self, widget: &P)`](http://gtk-rs.org/docs/gtk/trait.ContainerExt.html#tymethod.add). Now the method can be not only called with structs of that type, but also with structs of any subtype (in this case, any Widget).
```
GuillaumeGomez, line 17: "as arguments it pretty" -> "as arguments is pretty"

docs-src/tutorial/object_oriented.md
piegamesde: Argh, I hate it. "its" is so rare that my muscle memory always wants to write "it's". Already caught a few instances, missed that one.

GuillaumeGomez:
cc @sdroege @EPashkin

docs-src/tutorial/object_oriented.md
```diff
@@ -2,13 +2,23 @@
 layout: default
 ---
 
-# Upcast and downcast
+# Object-oriented Rust
 
-Since there is an inheritance system in Gtk, it's only logical to have one as well in `Gtk-rs`. Normally, most people won't need this, but it's always nice to understand how things work.
+Since there is an inheritance system in Gtk, it's only logical to have one as well in `Gtk-rs`. Normally, most people won't need this, but understanding how things work can help navigating the documentation a lot.
+
+Each GTK class is split into a struct named after that class, and a trait with the same name and the suffix `Ext`. All methods except the constructor(s) are not in the struct, but in that trait. **Notice 1:** some methods are hidden in a `…ExtManual` trait, as seen in [WidgetExtManual](http://gtk-rs.org/docs/gtk/prelude/trait.WidgetExtManual.html). **Notice 2:** The same principle applies to GTK interfaces as well (have a look at the [Orientable](http://gtk-rs.org/docs/gtk/struct.Orientable.html) interface for an example).
```
sdroege, line 9: There are also some types that don't have a trait and have every function on the struct (final types that can't have public subclasses).

sdroege:
Sounds good to me otherwise, thanks :)

docs-src/tutorial/object_oriented.md
piegamesde: When I wrote "GTK class", I was mainly thinking about the whole `Widget` type hierarchy. But yes, there are a few exceptions here and there. Should I somehow mention them or can I simply ignore that fact beacuse they are rare and most users shouldn't be bothered by this too much?

sdroege: There are a few widgets like that too. I would add a sentence about this so it doesn't come as a surprise. "An exception to this are final classes (which can't have further subclasses): they don't have an extension trait but define the functions directly on the type".

docs-src/tutorial/object_oriented.md
```diff
@@ -6,7 +6,7 @@ layout: default
...
+Each GTK class is split into a struct named after that class, and a trait with the same name and the suffix `Ext`. All methods except the constructor(s) are not in the struct, but in that trait. Some methods are hidden in a `…ExtManual` trait, as seen in [WidgetExtManual](http://gtk-rs.org/docs/gtk/prelude/trait.WidgetExtManual.html). The same principle applies to GTK interfaces as well (have a look at the [Orientable](http://gtk-rs.org/docs/gtk/struct.Orientable.html) interface for an example). There are a few exceptional classes that don't follow this general pattern: one reason for this is that final classes don't need an extension trait and thus can have their methods defined on the struct itself.
```
sdroege, line 9:
```suggestion
Each GTK class is split into a struct named after that class, and a trait with the same name and the suffix `Ext`, see for example [WidgetExt](http://gtk-rs.org/docs/gtk/prelude/trait.WidgetExt.html). All methods except the constructor(s) are not in the struct, but in that trait. In addition, some other methods can be found in a `…ExtManual` trait, as seen in [WidgetExtManual](http://gtk-rs.org/docs/gtk/prelude/trait.WidgetExtManual.html). This happens for implementation reasons currently. The same principle applies to GTK interfaces as well (have a look at the [Orientable](http://gtk-rs.org/docs/gtk/struct.Orientable.html) interface for an example). There are a few exceptional classes that don't follow this general pattern: one reason for this is that final classes don't need an extension trait because there can't be any subclasses of them, and thus can have their methods defined on the struct itself.
```

Maybe? Otherwise sounds good to me

piegamesde: Yes that's good, thank you.

GuillaumeGomez:
Thanks!

PR #159, opened 2020-07-05
Rename new_with_label to with_label

ryzokuken:
gtk::Button::new_with_label is no longer a thing. Replace calls to
new_with_label with with_label.


GuillaumeGomez:
Great catch, thanks!

PR #168, opened 2020-08-25
Added info about issue with Rust 2018

grelltrier:
When you create a new crate, cargo automatically sets the edition to 2018. This can cause issues with gir. I added a short info about the issue to the tutorial and how to avoid it.

### DIFF
--- a/docs-src/tutorial/gir_tutorial.md
+++ b/docs-src/tutorial/gir_tutorial.md
@@ -368,7 +368,9 @@ git = "https://github.com/gtk-rs/glib"
 git = "https://github.com/gtk-rs/sys" # all gtk-rs sys crates are in the sys repository
 ```
 
-And to import those crates into `src/lib.rs`:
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

docs-src/tutorial/gir_tutorial.md
```diff
@@ -368,7 +368,9 @@ git = "https://github.com/gtk-rs/glib"
 git = "https://github.com/gtk-rs/sys" # all gtk-rs sys crates are in the sys repository
 ```
 
-And to import those crates into `src/lib.rs`:
+While you are at it, you should also change the edition to 2015 or remove the field entirely from your Cargo.toml. This is advised because gir has an [issue](https://github.com/gtk-rs/gir/issues/746) with the 2018 edition of Rust. Don't worry, the created wrapper will work with the 2018 edition just fine.
+
+And to import the `libc` and `bitflags crates into `src/lib.rs`:
```
sdroege, line 373:
```suggestion
And to import the `libc` and `bitflags` crates into `src/lib.rs`:
```

grelltrier:
Gir now sets the edition to 2021 and this is obsolete

PR #170, opened 2020-08-25
Added info about global functions

grelltrier:
When generating bindings for their own project, users can have global functions. Added info on how to deal with them.

### DIFF
--- a/docs-src/tutorial/gir_tutorial.md
+++ b/docs-src/tutorial/gir_tutorial.md
@@ -315,7 +315,7 @@ Another `gir` run:
 ```
 
 There is also a third option you can encounter called [NOT GENERATED FUNCTION]. If you see this in your own project, it means you have global functions.
-To generate the code for all of them, add "YourGirName.*" to the generate = [ ] array. You also need to re-export them, so add 
+To generate the code for all of them, add "YourGirName.*" to the `generate = [ ]` array. You also need to re-export them, so add 
 `pub use self::auto::functions::*;` to your lib.rs file.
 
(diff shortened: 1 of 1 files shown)


docs-src/tutorial/gir_tutorial.md
```diff
@@ -314,6 +314,10 @@ Another `gir` run:
 > gir
 ```
 
+There is also a third option you can encounter called [NOT GENERATED FUNCTION]. If you see this in your own project, it means you have global functions.
+To generate the code for all of them, add "YourGirName.*" to the generate = [ ] array. You also need to re-export them, so add
```
sdroege, line 318:
```suggestion
To generate the code for all of them, add "YourGirName.*" to the `generate = [ ]` array. You also need to re-export them, so add 
```

docs-src/tutorial/gir_tutorial.md
```diff
@@ -314,6 +314,10 @@ Another `gir` run:
 > gir
 ```
 
+There is also a third option you can encounter called [NOT GENERATED FUNCTION]. If you see this in your own project, it means you have global functions.
```
sdroege, line 317:
```suggestion
There is also a third option you can encounter called `[NOT GENERATED FUNCTION]`. If you see this in your own project, it means you have global functions.
```

sdroege (approved):
Makes sense to me, apart from the formatting :) Thanks!

grelltrier:
This is also mentioned in the draft to rewrite the gir tutorial

PR #171, opened 2020-08-25
Added info about missing macros

grelltrier:
Sometimes gir generation fails because of missing macros. I added information on how that can be fixed.

### DIFF
--- a/docs-src/tutorial/gir_tutorial.md
+++ b/docs-src/tutorial/gir_tutorial.md
@@ -543,6 +543,29 @@ pub mod prelude;
 pub use prelude::*;
 ```
 
+### Cannot find macro
+If gir generation fails because it can't find one of the listed macros, it means you’ll have to implement them yourself. Depending on which macro is missing, you can add the following to your lib.rs file:
+
+/// No-op.
+macro_rules! skip_assert_initialized {
+    () => {};
+}
+
+/// Asserts that this is the main thread and either `gdk::init` or `gtk::init` has been called.
+macro_rules! assert_initialized_main_thread {
+    () => {
+        if !::gtk::is_initialized_main_thread() {
+            if ::gtk::is_initialized() {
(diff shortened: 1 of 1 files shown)


docs-src/tutorial/gir_tutorial.md
```diff
@@ -543,6 +543,29 @@ pub mod prelude;
 pub use prelude::*;
 ```
 
+### Cannot find macro
+If gir generation fails because it can't find one of the listed macros, it means you’ll have to implement them yourself. Depending on which macro is missing, you can add the following to your lib.rs file:
```
sdroege, line 547:
```suggestion
If compilation of the generated bindings fails because it can't find one of the listed macros, it means you’ll have to implement them yourself. Depending on which macro is missing, you can add the following to your `lib.rs` file:
```

docs-src/tutorial/gir_tutorial.md
```diff
@@ -543,6 +543,29 @@ pub mod prelude;
...
+            if ::gtk::is_initialized() {
+                panic!("GTK may only be used from the main thread.");
+            } else {
+                panic!("GTK has not been initialized. Call `gtk::init` first.");
+            }
+        }
+    };
+}
+
+One complication here is that the assert_initialized_main_thread macro depends on the exact library. If it's GTK-based then the above macro is likely correct, unless the library has its own initialization function (which would need to be handled in addition to GTK here). For non-GTK-based libraries this macro would do nothing if there's no library initialization function, and otherwise it handles that function.
```
sdroege, line 567:
```suggestion
One complication here is that the `assert_initialized_main_thread!` macro depends on the exact library. If it's GTK-based then the above macro is likely correct, unless the library has its own initialization function. If it has its own initialization function it would need to be handled in addition to GTK's here in the same way.
For non-GTK-based libraries this macro would handle the initialization function of that library in the same way, or if there is none would simply do nothing.
```

Issue #180, opened 2020-10-13
Cross-compilation guide partially outdated


anieuwland:
The very useful cross-compilation [guide](https://gtk-rs.org/docs-src/tutorial/cross) needs some extra instructions. Gettext [requires](https://docs.rs/crate/gettext-sys/0.19.9/source/build.rs) setting an environment variable pointing to the mingw installation. In the Compiling section, right before the build command, it is necessary to do `export MINGW_PREFIX=$GTK_INSTALL_PATH`.

Unfortunately, `GTK_INSTALL_PATH` is only introduced in a later section (Packaging). The guide might need a little restructuring to stay coherent. I also think it would be valuable to mention for example on Fedora the relevant path is `/usr/x86_64-w64-mingw32/sys-root/mingw/`. 

Lastly, I think *think* (but am not sure) there is a mistake in `export PKG_CONFIG_PATH=/usr/i686-w64-mingw32/lib/pkgconfig`. The whole guide is about x86_64, but here we point to i686 data, can that be right? 

So to summarize:

* The compiling section needs to mention setting MINGW_PREFIX to satisfy gettext. 
* It might mistakenly refer to i686 instead of x86_64 for mingw's pkg_config.
* The document might need a little rewriting to stay coherent and mention path equivalents from other distributions (notably Fedora).

What is the process to make this happen, if I'd like to?

sdroege:
You could create a PR that changes [this file](https://github.com/gtk-rs/gtk-rs.github.io/blob/master/docs-src/tutorial/cross.md). Thanks for noticing this :)

PR #181, opened 2020-10-15
Updated cross-compilation document

anieuwland:
I made the changes I outlined [here](https://github.com/gtk-rs/gtk-rs.github.io/issues/180) and also addressed the [issue](https://github.com/gtk-rs/gtk/issues/817) about the gtk3 libraries link being "bunk". Summarized:

* Added the MINGW_PREFIX instruction for a successful compile command
* Included Fedora instructions 
* Touched up capitalization and mark-up here and there
* Addressed the GTK link being dead by rewriting that section: now it gives the generic instructions and gives 3 options for acquiring the libraries, although none of them are ideal (recent libs downloadable with 1 click)

### DIFF
--- a/docs-src/tutorial/cross.md
+++ b/docs-src/tutorial/cross.md
@@ -4,7 +4,7 @@ layout: default
 
 # Preparing rust
 
-First install rust normally with rustup. The next step will install the windows toolchain.
+First install Rust normally with rustup. The next step will install the Windows toolchain.
 
     rustup target add x86_64-pc-windows-gnu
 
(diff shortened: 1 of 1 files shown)


sdroege:
Let's get this in then, reads good to me and is a definite improvement over the outdated information before. Too bad the situation sucks so much with GTK. It's much easier with GStreamer on Windows :)

anieuwland:
Definitely wish it were easier! But I'm happy to have been able to update these docs.

PR #182, opened 2020-11-08
Remove Way-Cooler from list

piegamesde:
It's a sad story, and I'm too lazy to find the link … simply look at their org, all rust repositories are deprecated.


sdroege:
Thanks! I remember that blog post, yes

PR #184, opened 2020-11-23
mention new gtk-rs application: projectpad

emmanueltouzery:
I wrote a new gtk-rs application (through relm), adding it to the list of gtk-rs applications.


GuillaumeGomez:
Nice!

PR #186, opened 2020-12-16
minor fix to the glade doc

plowpath:
while i was playing with gtk-rs noticed that ```new_``` has been deprecated (https://gtk-rs.org/blog/2020/07/04/new-release.html)


sdroege:
Thanks!

Issue #138, opened 2020-02-19
Request for additional information


windelbouwman:
I would like to request a few extra pages on the website.

A page on building on windows. I figured out how to build my app on windows with the help of a reddit post and some pieces of information in issues floating around, it would be great to have this information condensed on the website.

A page about resources and app deployment. This would be helpful as well, given the fact someone created an application, how is this best deployed to the various platforms? Topics such as resources (images, icons) could be described here.

Does this make sense?

GuillaumeGomez:
More documentation always make sense.

Firionus:
> A page on building on windows. I figured out how to build my app on windows with the help of a reddit post and some pieces of information in issues floating around, it would be great to have this information condensed on the website.

I'm a brand new user and documented my initial problems in https://github.com/gtk-rs/gtk-rs.github.io/issues/70#issuecomment-654752526.

sophie-h:
I think both those things are on the roadmap for the book @Hofer-Julian?

Hofer-Julian:
Yes.

PR #190, opened 2020-12-29
Document setup on Windows

### LINKED ISSUE #70: Add documentation about how to compile for Windows without cross-compilation
See https://www.reddit.com/r/rust/comments/86kmhu/compiling_rust_windows_gtk_stepbystep/ for a good start

jscissr:
As noted by others (#70, #138), and I also experienced myself, there is currently not enough documentation on using gtk-rs on Windows.

After multiple failed attempts, wondering what I was supposed to do and googling error messages, I found [this thread](https://github.com/gtk-rs/gtk/issues/44), which links to [this file](https://github.com/gtk-rs/examples/blob/master/appveyor.yml). I adapted and simplified these steps a bit, and can confirm that they actually work.

Let's avoid every newcomer wasting a lot of time by adding this to the documentation.

Fixes #70

### DIFF
--- a/docs-src/index.md
+++ b/docs-src/index.md
@@ -7,6 +7,24 @@ layout: default
 
 First, prepare your system by taking a look at the [GTK installation page](https://www.gtk.org/docs/installations/).
 
+### Setup on Windows
+
+On Windows, some Rust-specific steps are necessary:
+
+1. If you haven't already, [install rustup](https://rustup.rs/)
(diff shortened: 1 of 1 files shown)


categulario:
I just recently had this same problem and your addition would have helped me but it also assumes the windows-gnu toolchain which is not always necessary what you want. For instance I solved my problems using the msvc toolchain.

Perhaps this topic requires its own page explaining the possibilities. (And I would happily contribute to it).

Issue #192, opened 2021-01-07
Issue in https://gtk-rs.org/docs-src/tutorial/cross


InfRandomness:
Hi,

it looks like there is an issue in the cross.md file displayed by https://gtk-rs.org/docs-src/tutorial/cross at the openSUSE part,

it currently shows like that on two different computers, both using firefox 84.0.1
![image](https://user-images.githubusercontent.com/43730933/103955541-c9322780-5146-11eb-8f65-fe2bd2629377.png)

this doesn't match the content in the md file at all,
when visualizing the same file at https://github.com/gtk-rs/gtk-rs.github.io/blob/master/docs-src/tutorial/cross.md, everything is displayed correctly though

sdroege:
@GuillaumeGomez Any ideas?

GuillaumeGomez:
Content not updated yet on github side? No idea otherwise...

InfRandomness:
Well.. this is kinda' weird, the content that's being shown doesn't match any of the two commits I made https://github.com/gtk-rs/gtk-rs.github.io/pull/191

GuillaumeGomez:
Yep, but no idea what's going on... I don't have control over the "publication" of the content unfortunately...

InfRandomness:
That's really weird..

GuillaumeGomez:
Maybe ask the github staff directly? I have no idea otherwise...

InfRandomness:
What's even weirder, as I've just noticed is that at the line 28 "### Fedora" doesn't even seems to show up either ? But at line 19, "### Arch Linux" shows up, and I'm wondering if that's not because there's a carriage return between the line 19 and 17, which is not the case for the line 28 that contains "### Fedora" and for the line 33 "### openSUSE"

InfRandomness:
I really don't think the issue is related to the hosting solution but I think it is related to how the website treats the md file

GuillaumeGomez:
Ok, let me double-check the markdown and try to understand what's going on.

InfRandomness:
This confirms my saying, I added carriage returns between lines containing '### ...' and other lines and run the website locally.

PR #193, opened 2021-01-08
Fix Heading issues

InfRandomness:
Fixes https://github.com/gtk-rs/gtk-rs.github.io/issues/192


GuillaumeGomez (approved):
Ah indeed. I remember wanting to nitpick about that the missing newlines but didn't because I was already quite happy of the contribution. Well, now we know that they're quite important. Thanks a lot!

PR #197, opened 2021-03-07
Add `mingw64-winpthreads` in Fedora dependencies

adnan449:
As observed and suggested in https://github.com/gtk-rs/gtk-rs/issues/187#issuecomment-751454832


GuillaumeGomez:
Thanks!

PR #224, opened 2021-06-13
Fix doc icon for temporary v0.9 manual crates list

### LINKED ISSUE #223: Doc images are broken
Screenshot from 2021-06-13 14-12-24

cc @sophie-h

sophie-h:
Original fix is in #217, but was not applied to the temporary manual crates list.

image

Closes #223


index.md
```diff
@@ -55,7 +55,14 @@ The following table contains the most popular crates of **gtk-rs**. More informa
       v0.9.0
     </td>
     <td class="docs">
-      <a href="/gtk-rs-core/stable/latest/docs/cairo">🕮</a>
+      <a href="/gtk-rs-core/stable/latest/docs/cairo">
+        <?xml version="1.0" encoding="UTF-8"?>
+<svg width="181" height="105" version="1.1" viewBox="0 0 47.89 27.781" xmlns="http://www.w3.org/2000/svg" xmlns:cc="http://creativecommons.org/ns#" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
```
GuillaumeGomez, line 60: Instead of redeclaring the image multiple times, couldn't it be stored in a file instead?

GuillaumeGomez: (Even if not redeclared, it's very hard to read anything, it's supposed to be a markdown file)

index.md
```diff
@@ -36,7 +36,7 @@ The following table contains the most popular crates of **gtk-rs**. More informa
 {% include crates.html %}
 -->
 
-<table>
+  <table class="crates">
```
GuillaumeGomez, line 39: Please remove this indent.

index.md
sophie-h: This is a temporary (!!!) fix because we are still lacking the v0.9 docs for gtk3 in the new system (cc @bilelmoussaoui). The correct solution would just be `{% include crates.html %}` which generates it from a json file. But that would lead to incorrect docs links for now.

bilelmoussaoui: ah i always forget about those. Let me fix them :)

GuillaumeGomez: If you mean by that that we will generate the svg images directly into the HTML, please don't and create an image file instead.

sophie-h: Then you have to fix it yourself. Coloring with filters seems like a complicated detour. I think there is also a way with iframes? But I never used it.

sophie-h: You could also generate all svgs in the page with the required colors of course.

PR #229, opened 2021-06-17
Add featured projects and ecosystem section

sophie-h:
- Featured Projects: Based on the "slick" carousel. This is also used
  by gtk.org.
- Ecosystem: Link to gstreamer-rs, libhandy-rs, libadwaita-rs.

##### Todo

- [ ] Agree on featured apps
- [ ] Better screenshots

Closes #213, #208

PR #234, opened 2021-06-19
Featured projects

GuillaumeGomez:
This is based on https://github.com/gtk-rs/gtk-rs.github.io/pull/229.

@sophie-h: I only kept the slick font (for the slider arrows).


_sass/_base.scss
```diff
@@ -1,3 +1,12 @@
+@font-face {
+    font-family: 'slick';
+    font-weight: normal;
+    font-style: normal;
+
+    src: url('/font/slick.eot');
+    src: url('/font/slick.eot?#iefix') format('embedded-opentype'), url('/font/slick.woff') format('woff'), url('/font/slick.ttf') format('truetype'), url('/font/slick.svg#slick') format('svg');
```
sophie-h, line 7: I think the font itself is missing from the commit. And probably one format it enough? (I'm servering Cantarell only as woff.)

_sass/_layout.scss
```diff
@@ -269,13 +291,78 @@ ul.contact {
     line-height: 1;
 }
 
+/**
+ * Carousel
```
sophie-h, line 295: Is this still in use?

sophie-h (commented):
There seems to be something of with the swipe. At the ends: To the left it jumps to the second. To the right it moves to empty page. Buttons work though.

sophie-h:
This looks very nice!

_sass/_base.scss
sophie-h: Yes, research says: just keep the woff.

_sass/_layout.scss
GuillaumeGomez: No indeed! Removed.

_sass/_base.scss
GuillaumeGomez: Done. :)

GuillaumeGomez:
So if we all agree with this, I'll squash commits so the extra files don't appear in the history.

sophie-h:
We will have to check what to do with the screenshot size if we reduce the content width. But imo let's merge this first.

GuillaumeGomez:
It's handled quite nicely. :)

GuillaumeGomez:
Ok, time to test now if I didn't break everything. ;)

PR #236, opened 2021-06-20
Fix slide display issues

GuillaumeGomez:
* The buttons are now hidden on mobile.
* The slider is more "permissive" (only need to slide 30% to go to the next one).
* Fix unwanted whitespaces between slides.


GuillaumeGomez:
Gonna go ahead and merge it so people can test it. :)

PR #247, opened 2021-06-25
Fallback to monospace for code

nagisa:
Certain configurations of Firefox allow disabling custom web fonts. In
those instances the code samples are using a sans-serif font (if `Roboto
Mono` is not otherwise available).


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

cc @sophie-h

PR #249, opened 2021-06-27
2021 gtk-rs release post: cairo doesn't use any traits

bilelmoussaoui:
cairo itself doesn't use prelude and it doesn't make sense to include it in the prelude example

### DIFF
--- a/_posts/2021-06-22-new-release.md
+++ b/_posts/2021-06-22-new-release.md
@@ -92,7 +92,7 @@ in the `prelude`, so importing it will give you access to them:
 use gtk::prelude::*;
 
 // ...
-let x = cairo::something();
+let x = pango::something();
 let y = gdk::something();
 ```
 


_posts/2021-06-22-new-release.md
```diff
@@ -92,7 +92,7 @@ in the `prelude`, so importing it will give you access to them:
 use gtk::prelude::*;
```
piegamesde, line 93: Um, isn't there a `use gtk::pango` missing if one does not declare the `pango` dependency separately?

bilelmoussaoui: the example is supposed to tell about not having to include the traits manually from the various crates.

bilelmoussaoui: although, i agree with your concern, the example is just not great

sdroege: Let's fix the example then:

```rust

let something = gtk::SomeThing::new();
let something_else = something.pango_layout();
pango_layout.do_things();
```

Exercise for someone knowing GTK to look up some API from where you'd need this :)

PR #264, opened 2022-01-14
[release] merging release-2022-01-14 into master

GuillaumeGomez:
cc @GuillaumeGomez @sdroege @bilelmoussaoui


_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Add LICENSE files in sys crates](https://github.com/gtk-rs/gtk-rs-core/pull/173)
+ * [Add LICENSE files in sys crates](https://github.com/gtk-rs/gtk-rs-core/pull/174)
+ * [docs: handle minor releases](https://github.com/gtk-rs/gtk-rs-core/pull/176)
+ * [Fix compiler errors in the tests related to c\_char signedness differe…](https://github.com/gtk-rs/gtk-rs-core/pull/175)
+ * [backport: handle minor releases docs](https://github.com/gtk-rs/gtk-rs-core/pull/177)
+ * [Fix compiler errors in the tests related to c\_char signedness differe…](https://github.com/gtk-rs/gtk-rs-core/pull/178)
+ * [Run clippy on beta](https://github.com/gtk-rs/gtk-rs-core/pull/182)
+ * [Remove unwanted warnings from the clone macro](https://github.com/gtk-rs/gtk-rs-core/pull/183)
+ * [Remove unwanted warnings from the clone macro](https://github.com/gtk-rs/gtk-rs-core/pull/185)
+ * [Make stronger guarantees on `ToValue`](https://github.com/gtk-rs/gtk-rs-core/pull/188)
```
sdroege, line 40: This is worth mentioning

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Add cairo::Result type alias](https://github.com/gtk-rs/gtk-rs-core/pull/184)
+ * [Update versions to 0.14.1](https://github.com/gtk-rs/gtk-rs-core/pull/191)
+ * [0.14 backports](https://github.com/gtk-rs/gtk-rs-core/pull/190)
+ * [Replace unneeded doc comments with comments](https://github.com/gtk-rs/gtk-rs-core/pull/192)
+ * [glib: Re-export main context acquire guard](https://github.com/gtk-rs/gtk-rs-core/pull/194)
+ * [glib: Re-export main context acquire guard](https://github.com/gtk-rs/gtk-rs-core/pull/195)
+ * [Update versions to 0.14.2](https://github.com/gtk-rs/gtk-rs-core/pull/196)
+ * [Bump min required version to 1.53.0](https://github.com/gtk-rs/gtk-rs-core/pull/210)
```
sdroege, line 48: This should be mentioned, but our minimum is now 1.56

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Update gir for new builder `#\[must\_use\]` usage](https://github.com/gtk-rs/gtk3-rs/pull/687)
+ * [gtk: manually implement gtk\_print\_operation\_get\_error](https://github.com/gtk-rs/gtk3-rs/pull/689)
+ * [Regenerate with latest gir (assert safe gerror)](https://github.com/gtk-rs/gtk3-rs/pull/688)
+ * [gtk/builder: manually implement several methods (custom return codes)](https://github.com/gtk-rs/gtk3-rs/pull/691)
+ * [Regenerate with latest gir (gboolean return codes)](https://github.com/gtk-rs/gtk3-rs/pull/692)
+ * [Remove unused heck dependency and use system-deps 6 for gdkwayland too](https://github.com/gtk-rs/gtk3-rs/pull/693)
+ * [Rename `impl\_` to `imp`](https://github.com/gtk-rs/gtk3-rs/pull/694)
+ * [Auto generate with fixed slice of copy types](https://github.com/gtk-rs/gtk3-rs/pull/695)
+ * [\[gir\] Generate #\[must\_use\] where needed](https://github.com/gtk-rs/gtk3-rs/pull/697)
+ * [regenerate with latest gir-files](https://github.com/gtk-rs/gtk3-rs/pull/698)
```
sdroege, line 311: Should mention somewhere central what the minimum/maximum versions of the C libraries are that we support at this point

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Update versions to 0.14.2](https://github.com/gtk-rs/gtk-rs-core/pull/196)
+ * [Bump min required version to 1.53.0](https://github.com/gtk-rs/gtk-rs-core/pull/210)
+ * [cairo: silence self\_named\_constructor](https://github.com/gtk-rs/gtk-rs-core/pull/217)
+ * [glib: ignore unref return value](https://github.com/gtk-rs/gtk-rs-core/pull/215)
+ * [Added basic gio::Initable implementation](https://github.com/gtk-rs/gtk-rs-core/pull/218)
+ * [add pango\_attr\_font\_desc\_new](https://github.com/gtk-rs/gtk-rs-core/pull/222)
+ * [Merge README files and crate documentation](https://github.com/gtk-rs/gtk-rs-core/pull/219)
+ * [Variant,VariantDict: More type lookup methods](https://github.com/gtk-rs/gtk-rs-core/pull/216)
+ * [be able to inspect pango attribute types and downcast them](https://github.com/gtk-rs/gtk-rs-core/pull/221)
+ * [Add function to consume raw data of ImageSurface](https://github.com/gtk-rs/gtk-rs-core/pull/220)
```
sdroege, line 56: This seems like a useful helper function

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [glib/variant: Expose `g\_variant\_get\_normal\_form()`](https://github.com/gtk-rs/gtk-rs-core/pull/223)
+ * [glib/variant: add fallible child-value getter](https://github.com/gtk-rs/gtk-rs-core/pull/224)
+ * [variant: Use into\_owned()](https://github.com/gtk-rs/gtk-rs-core/pull/228)
+ * [variant: Add more APIs to access typed children](https://github.com/gtk-rs/gtk-rs-core/pull/226)
+ * [variant: Add a `bytes()` method.](https://github.com/gtk-rs/gtk-rs-core/pull/227)
+ * [variant: Add an API to iterate over `as`](https://github.com/gtk-rs/gtk-rs-core/pull/229)
```
sdroege, line 62: Should get a section about all the variant API improvements including the derive macro

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Implement FromGlibPtrBorrow for glib::Value and glib::SendValue](https://github.com/gtk-rs/gtk-rs-core/pull/275)
+ * [Update versions](https://github.com/gtk-rs/gtk-rs-core/pull/277)
+ * [Update gir with removed type version heuristics](https://github.com/gtk-rs/gtk-rs-core/pull/278)
+ * [docs: Fix the examples for variants](https://github.com/gtk-rs/gtk-rs-core/pull/279)
+ * [glib: Merge variants of `wrapper!` as much as possible](https://github.com/gtk-rs/gtk-rs-core/pull/280)
+ * [Use bool instead of gboolean in graphene for FFI](https://github.com/gtk-rs/gtk-rs-core/pull/282)
+ * [0.14 backports](https://github.com/gtk-rs/gtk-rs-core/pull/283)
+ * [Clean up source removal](https://github.com/gtk-rs/gtk-rs-core/pull/286)
+ * [Remove unused dependencies](https://github.com/gtk-rs/gtk-rs-core/pull/288)
+ * [Add support for inline-allocated Boxed types and various graphene fixups](https://github.com/gtk-rs/gtk-rs-core/pull/284)
```
sdroege, line 99: This is worth mentioning as it reduces allocations for various things

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [value: Don't take internal value out of the Value struct but instead …](https://github.com/gtk-rs/gtk-rs-core/pull/290)
+ * [cairo: remove target\_os = "ios"](https://github.com/gtk-rs/gtk-rs-core/pull/293)
+ * [Remove iOS target from quartz\_surface in Cairo](https://github.com/gtk-rs/gtk-rs-core/pull/292)
+ * [Fix new clippy warnings](https://github.com/gtk-rs/gtk-rs-core/pull/296)
+ * [Include the LICENSE file in the published crates](https://github.com/gtk-rs/gtk-rs-core/pull/298)
+ * [glib: Implement TimeSpan as a wrapper type with some convenience API](https://github.com/gtk-rs/gtk-rs-core/pull/299)
+ * [Declare minimum supported Rust version in Cargo.toml](https://github.com/gtk-rs/gtk-rs-core/pull/295)
+ * [Main context fixups](https://github.com/gtk-rs/gtk-rs-core/pull/300)
```
sdroege, line 107: This should be mentioned

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [glib: Provide default implementations for `IsSubclassable` methods](https://github.com/gtk-rs/gtk-rs-core/pull/256)
+ * [Allow to change visibility in glib::wrapper macro](https://github.com/gtk-rs/gtk-rs-core/pull/304)
+ * [Add missing pango attrs & cleanup glyph\*](https://github.com/gtk-rs/gtk-rs-core/pull/308)
+ * [Various Pango fixes](https://github.com/gtk-rs/gtk-rs-core/pull/305)
+ * [Make parameters to `gio::ListStore::splice()` more generic](https://github.com/gtk-rs/gtk-rs-core/pull/309)
+ * [glib: impl FromGlibContainerAsVec for \*const $ffi\_name as well for Bo…](https://github.com/gtk-rs/gtk-rs-core/pull/311)
+ * [Minor cleanup as preparation for edition 2021](https://github.com/gtk-rs/gtk-rs-core/pull/312)
+ * [pango: implement DerefMut for Attribute](https://github.com/gtk-rs/gtk-rs-core/pull/313)
+ * [Edition 2021](https://github.com/gtk-rs/gtk-rs-core/pull/314)
+ * [glib: add simpler variants of ParamSpec::new\_override](https://github.com/gtk-rs/gtk-rs-core/pull/315)
```
sdroege, line 117: Maybe @bilelmoussaoui wants to mention this

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [object: fix documentation typo](https://github.com/gtk-rs/gtk-rs-core/pull/322)
+ * [gio: Add various `gio::File` missing functions](https://github.com/gtk-rs/gtk-rs-core/pull/319)
+ * [gio: add manual bindings for GIOExtension](https://github.com/gtk-rs/gtk-rs-core/pull/316)
+ * [Backport IOExtensionPoint bindings to 0.14](https://github.com/gtk-rs/gtk-rs-core/pull/324)
+ * [None constants as associated constants](https://github.com/gtk-rs/gtk-rs-core/pull/327)
+ * [glib: Assert that the memory allocated by GLib for the private instan…](https://github.com/gtk-rs/gtk-rs-core/pull/328)
+ * [Add missing Matrix methods](https://github.com/gtk-rs/gtk-rs-core/pull/330)
+ * [glib: add a Value::get\_owned](https://github.com/gtk-rs/gtk-rs-core/pull/325)
+ * [gio: Make progress callback in `File::copy\_async` optional](https://github.com/gtk-rs/gtk-rs-core/pull/331)
+ * [glib: Remove `glib::String` bindings](https://github.com/gtk-rs/gtk-rs-core/pull/332)
```
sdroege, line 132: Mention that this became `glib::StringBuilder` with better API

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Remove Into&lt;&str&gt; trait bounds and directly use &str](https://github.com/gtk-rs/gtk-rs-core/pull/333)
+ * [variant: Add support for `BTreeMap`](https://github.com/gtk-rs/gtk-rs-core/pull/334)
+ * [ObjectExt helpers](https://github.com/gtk-rs/gtk-rs-core/pull/323)
```
sdroege, line 135: More convenient functions that avoid all the `unwrap`s for the common case from @bilelmoussaoui

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Some more `Variant` API additions](https://github.com/gtk-rs/gtk-rs-core/pull/344)
+ * [glib: Fix `Deref` impls for `FixedSizeVariantArray`](https://github.com/gtk-rs/gtk-rs-core/pull/347)
+ * [glib: Mark wrapper structs for `Shared`/`Object`/`Boxed`/`Type` as `#\[repr(transparent)\]`](https://github.com/gtk-rs/gtk-rs-core/pull/348)
+ * [macros: implement FromGlibNone/Full for shared boxed types.](https://github.com/gtk-rs/gtk-rs-core/pull/346)
+ * [Fix #335 - repr(transparent) for cairo\_bool\_t; correct opaque types](https://github.com/gtk-rs/gtk-rs-core/pull/336)
+ * [glib: Implement Send/Sync/PartialOrd/Ord/Hash on GPtrSlice/GSlice](https://github.com/gtk-rs/gtk-rs-core/pull/349)
+ * [glib: Re-use paramspec names instead of creating a new NUL-terminated…](https://github.com/gtk-rs/gtk-rs-core/pull/351)
+ * [glib: Simplify enum/flags class implementation](https://github.com/gtk-rs/gtk-rs-core/pull/352)
+ * [pango: take a generic attr in AttrList methods](https://github.com/gtk-rs/gtk-rs-core/pull/354)
+ * [macros: make ParentType optional](https://github.com/gtk-rs/gtk-rs-core/pull/356)
```
sdroege, line 149: Worth mentioning or not?

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [ regen (correctly generate opaque types)](https://github.com/gtk-rs/gtk-rs-core/pull/353)
+ * [Add wrapper type around `GList` / `GSList` and various container cleanups/improvements](https://github.com/gtk-rs/gtk-rs-core/pull/355)
+ * [Remove `glib::Array` bindings and make `glib::ByteArray` bindings safer (plus micro-optimization)](https://github.com/gtk-rs/gtk-rs-core/pull/358)
+ * [gio: don't expose task module](https://github.com/gtk-rs/gtk-rs-core/pull/360)
+ * [pango cleanups](https://github.com/gtk-rs/gtk-rs-core/pull/357)
+ * [gdk-pixbuf: generate missing auto bindings](https://github.com/gtk-rs/gtk-rs-core/pull/361)
+ * [Add gclosure macro](https://github.com/gtk-rs/gtk-rs-core/pull/338)
```
sdroege, line 156: This should probably mentioned

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Remove Into&lt;&str&gt; trait bounds and directly use &str](https://github.com/gtk-rs/gtk-rs-core/pull/333)
+ * [variant: Add support for `BTreeMap`](https://github.com/gtk-rs/gtk-rs-core/pull/334)
+ * [ObjectExt helpers](https://github.com/gtk-rs/gtk-rs-core/pull/323)
+ * [error: make message() public](https://github.com/gtk-rs/gtk-rs-core/pull/339)
+ * [glib: Lots of `Variant` / `VariantType` improvements](https://github.com/gtk-rs/gtk-rs-core/pull/340)
+ * [glib-macros: Add a `Variant` trait derive macro](https://github.com/gtk-rs/gtk-rs-core/pull/341)
+ * [re-generate with per crate min version](https://github.com/gtk-rs/gtk-rs-core/pull/345)
+ * [Some more `Variant` API additions](https://github.com/gtk-rs/gtk-rs-core/pull/344)
+ * [glib: Fix `Deref` impls for `FixedSizeVariantArray`](https://github.com/gtk-rs/gtk-rs-core/pull/347)
+ * [glib: Mark wrapper structs for `Shared`/`Object`/`Boxed`/`Type` as `#\[repr(transparent)\]`](https://github.com/gtk-rs/gtk-rs-core/pull/348)
```
sdroege, line 142: This is worth a note because it basically means that all our wrapper types are exactly the C/FFI raw pointer now. Which allows for all kinds of nice optimizations.

sdroege: There were also some more changes related to signal emissions IIRC

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Update macro syntax for enum, flags, boxed, and variant](https://github.com/gtk-rs/gtk-rs-core/pull/364)
+ * [glib: fix broken link](https://github.com/gtk-rs/gtk-rs-core/pull/391)
+ * [pango: opt-in for gi-docgen](https://github.com/gtk-rs/gtk-rs-core/pull/392)
+ * [glib: return a ParamSpec for override methods](https://github.com/gtk-rs/gtk-rs-core/pull/393)
+ * [glib: add missing doc-aliases to Type](https://github.com/gtk-rs/gtk-rs-core/pull/395)
+ * [Fix various 1.58 (beta) clippy warnings or silence them](https://github.com/gtk-rs/gtk-rs-core/pull/396)
+ * [Generate missing const doc aliases](https://github.com/gtk-rs/gtk-rs-core/pull/397)
+ * [glib: Add Gir.toml check and fix order in Gir.toml](https://github.com/gtk-rs/gtk-rs-core/pull/398)
+ * [Update gir for new builder `#\[must\_use\]` usage](https://github.com/gtk-rs/gtk-rs-core/pull/399)
+ * [glib: make spawn\*() return the SourceId](https://github.com/gtk-rs/gtk-rs-core/pull/400)
```
sdroege, line 182: This is worth a note together with cancellation of futures in general

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [graphene: add getters/setters for Size/Point/Point3D](https://github.com/gtk-rs/gtk-rs-core/pull/389)
+ * [glib: add ObjectSubclassIsExt trait with impl\_ method](https://github.com/gtk-rs/gtk-rs-core/pull/405)
+ * [glib: add ObjectSubclassIsExt to the prelude](https://github.com/gtk-rs/gtk-rs-core/pull/406)
+ * [Regenerate with trimmed extra \_async](https://github.com/gtk-rs/gtk-rs-core/pull/407)
+ * [Regenerate with latest gir (gboolean return codes)](https://github.com/gtk-rs/gtk-rs-core/pull/414)
+ * [allow for the modifying GlyphInfo and GlyphGeometry](https://github.com/gtk-rs/gtk-rs-core/pull/413)
+ * [docs: add missing rustdoc-stripper-ignore-next & test that on CI](https://github.com/gtk-rs/gtk-rs-core/pull/418)
+ * [Implement `ValueTypeOptional` for shared/boxed/object types in `glib:…](https://github.com/gtk-rs/gtk-rs-core/pull/420)
+ * [glib-macros: Update to heck 0.4](https://github.com/gtk-rs/gtk-rs-core/pull/425)
+ * [glib: add AnyGObject](https://github.com/gtk-rs/gtk-rs-core/pull/410)
```
sdroege, line 196: Mention or not? How useful is that in gtk @bilelmoussaoui ?

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [examples: Don't silently ignore errors when property bindings can't b…](https://github.com/gtk-rs/gtk3-rs/pull/659)
+ * [gdk: bind WaylandWindow manually](https://github.com/gtk-rs/gtk3-rs/pull/660)
+ * [implement subclassing for scrolledwindow](https://github.com/gtk-rs/gtk3-rs/pull/661)
+ * [None constants as associated constants](https://github.com/gtk-rs/gtk3-rs/pull/662)
+ * [Update per ObjectExt changes](https://github.com/gtk-rs/gtk3-rs/pull/663)
+ * [re-generate with per crate min version](https://github.com/gtk-rs/gtk3-rs/pull/668)
+ * [Regen (prepend `Widget`)](https://github.com/gtk-rs/gtk3-rs/pull/669)
+ * [examples: update per pango::AttrList simplification](https://github.com/gtk-rs/gtk3-rs/pull/671)
+ * [regen (correctly generate opaque types)](https://github.com/gtk-rs/gtk3-rs/pull/670)
+ * [Add missing wayland types and create new gdkwayland crate](https://github.com/gtk-rs/gtk3-rs/pull/665)
```
sdroege, line 287: This by @GuillaumeGomez

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [gtk: manually implement gtk\_print\_operation\_get\_error](https://github.com/gtk-rs/gtk3-rs/pull/689)
+ * [Regenerate with latest gir (assert safe gerror)](https://github.com/gtk-rs/gtk3-rs/pull/688)
+ * [gtk/builder: manually implement several methods (custom return codes)](https://github.com/gtk-rs/gtk3-rs/pull/691)
+ * [Regenerate with latest gir (gboolean return codes)](https://github.com/gtk-rs/gtk3-rs/pull/692)
+ * [Remove unused heck dependency and use system-deps 6 for gdkwayland too](https://github.com/gtk-rs/gtk3-rs/pull/693)
+ * [Rename `impl\_` to `imp`](https://github.com/gtk-rs/gtk3-rs/pull/694)
+ * [Auto generate with fixed slice of copy types](https://github.com/gtk-rs/gtk3-rs/pull/695)
+ * [\[gir\] Generate #\[must\_use\] where needed](https://github.com/gtk-rs/gtk3-rs/pull/697)
+ * [regenerate with latest gir-files](https://github.com/gtk-rs/gtk3-rs/pull/698)
+ * [manually implement Container::(get|set)\_property](https://github.com/gtk-rs/gtk3-rs/pull/699)
```
sdroege, line 312: This might be useful. Generic child property setter/getter

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [Update for glib wrapper API changes](https://github.com/gtk-rs/gtk3-rs/pull/701)
+ * [Update gir files](https://github.com/gtk-rs/gtk3-rs/pull/702)
+ * [ci: Allow explicitly triggering CI runs](https://github.com/gtk-rs/gtk3-rs/pull/703)
+ * [Fix new clippy lints](https://github.com/gtk-rs/gtk3-rs/pull/704)
+
+[gtk4-rs](https://github.com/gtk-rs/gtk4-rs):
```
sdroege, line 318: I'll leave this to @bilelmoussaoui

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [glib: Fix missing indirection in previous commit and add tests](https://github.com/gtk-rs/gtk-rs-core/pull/377)
+ * [Update to system-deps 6](https://github.com/gtk-rs/gtk-rs-core/pull/378)
+ * [glib: Use plain FFI functions instead of a GClosure for the property …](https://github.com/gtk-rs/gtk-rs-core/pull/381)
+ * [glib: Validate signal class handler return value and arguments when chaining up](https://github.com/gtk-rs/gtk-rs-core/pull/380)
+ * [glib: Mark `g\_get\_user\_special\_dir()` return value as `nullable`](https://github.com/gtk-rs/gtk-rs-core/pull/383)
+ * [Rename various constructors for consistency](https://github.com/gtk-rs/gtk-rs-core/pull/384)
+ * [glib: Fix FromValue impls for &Boxed, &Shared and &Object](https://github.com/gtk-rs/gtk-rs-core/pull/385)
+ * [glib: Add API for non-destructive iteration over a `List` / `SList`](https://github.com/gtk-rs/gtk-rs-core/pull/386)
+ * [glib: Add new trait for allowing to unwrap the return value of `Closu…](https://github.com/gtk-rs/gtk-rs-core/pull/366)
+ * [Update macro syntax for enum, flags, boxed, and variant](https://github.com/gtk-rs/gtk-rs-core/pull/364)
```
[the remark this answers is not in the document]
sdroege: > Maybe talk in this para that we now have a closure macro

I'd mention that in a separate section and talk about how to use that for signals

pbor:
I am not sure if we want to mention this, since I do not think gio bindings are very used, but if we decide we want it, we can add a para along the lines of:

gio bindings saw many additions and improvements, among other things it is now possible to implement the `gio::Initable` interface, the `gio::IOExtention` api is now available, the `gio::Task` bindings have been reworked to address soundness issues

_posts/2022-01-14-new-release.md
sdroege:
```markdown
With this release the minimum supported Rust version was updated to 1.56. At the same time all crates were updated to the Rust 2021 edition but backwards compatibility with the Rust 2015 edition is kept and crates can use the gtk-rs macros while staying with older editions.
```

_posts/2022-01-14-new-release.md
sdroege:
```markdown
The minimum supported versions of the underlying C libraries have not changed compared to last release but support for newer versions was added.
 * GLib/GObject/GIO: minimum 2.48, newest 2.72
 * Pango/PangoCairo: minimum 1.38, newest 1.52
 * gdk-pixbuf: minimum 2.32, newest 2.42
 * cairo: minimum 1.14, newest 1.16
 * graphene: minimum 1.10
 * ATK: minimum 2.18, newest 2.38
 * GTK3: minimum 3.18, newest, 3.24.9
 * GTK4: minimum 4.0, newest 4.6
```

_posts/2022-01-14-new-release.md
sdroege:
```markdown
Cairo image surfaces can now be converted to an owned value that behaves like an `&mut [u8]` and that can be safely sent to different threads. This allows both sending Cairo image surfaces to other threads safely as well as their underlying data, e.g. to fill an image surface from a compressed image via the `image` crate or saving the contents of an image surface to compressed data from a different thread.

```rust
let surface =  cairo::ImageSurface::create(cairo::Format:ARgb32, 320, 240).except("Failed to allocate memory");
// Use Cairo API on the `surface`.
// [...]
let data = surface.take_data().expect("Not the unique owner of the surface");
// Send `data` to another thread or otherwise treat it as a `&mut [u8]`.
// [...]
let surface = data.into_inner();
// Use the surface with Cairo APIs again.
```

This solves one of the common problems when using Cairo in multi-threaded applications.
```

File: _posts/2022-01-14-new-release.md (hunk 5 above)
Comment (sdroege) [reply _posts/2022-01-14-new-release.md:62]:
```markdown
The [`glib::Variant`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/variant/struct.Variant.html) API was improved a lot in this release. `Variant` is a type that allows to store structured data built from basic types in a way that can be serialized into binary data efficiently.

A lot of API that was missing in previous releases was added, including API for efficient conversion from/to slices of basic types, strings and other container types. The existing conversion APIs were optimized and cleaned up.

To make it easier to handle custom structs (and soon [enums](https://github.com/gtk-rs/gtk-rs-core/pull/434)), a [derive macro](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/derive.Variant.html) was added that allows to directly convert between a custom Rust type and a `Variant` and back.

```rust
use glib::prelude::*;

#[derive(Debug, PartialEq, Eq, glib::Variant)]
struct Foo {
    some_string: String,
    some_int: i32,
}

let v = Foo { some_string: String::from("bar"), some_int: 1 };
let var = v.to_variant();
assert_eq!(var.get::<Foo>(), Some(v));
```

This can be helpful for sending structured data over DBus via the `gio` DBus API, or for dealing with [`gio::Settings`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/gio/struct.Settings.html) or for more complex states/values used in [`gio::Action`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/gio/struct.Action.html).

```

File: _posts/2022-01-14-new-release.md (hunk 6 above)
Comment (sdroege) [reply _posts/2022-01-14-new-release.md:99]:
```markdown
Various types that are usually used stack-allocated from C were previously heap-allocated in the Rust bindings. With this release, these values are also stack-allocated when used from Rust unless explicitly heap-allocated, e.g. by explicitly placing them into a `Box`.

Examples for this are all the [`graphene`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/graphene/index.html) types (vectors, matrices, etc), which should make calculations considerably more efficient, types like [`gdk::Rectangle`](https://gtk-rs.org/gtk4-rs/stable/0.15/docs/gdk4/struct.Rectangle.html) and [`gdk::RGBA`](https://gtk-rs.org/gtk4-rs/stable/0.15/docs/gdk4/struct.RGBA.html), and [`gtk::TreeIter`](https://gtk-rs.org/gtk4-rs/stable/0.15/docs/gtk4/struct.TreeIter.html).

Overall this brings performance of applications using the Rust bindings even closer to the same code in C.
```

_posts/2022-01-14-new-release.md
```diff
@@ -0,0 +1,600 @@
...
+ * [error: make message() public](https://github.com/gtk-rs/gtk-rs-core/pull/339)
+ * [glib: Lots of `Variant` / `VariantType` improvements](https://github.com/gtk-rs/gtk-rs-core/pull/340)
+ * [glib-macros: Add a `Variant` trait derive macro](https://github.com/gtk-rs/gtk-rs-core/pull/341)
+ * [re-generate with per crate min version](https://github.com/gtk-rs/gtk-rs-core/pull/345)
+ * [Some more `Variant` API additions](https://github.com/gtk-rs/gtk-rs-core/pull/344)
+ * [glib: Fix `Deref` impls for `FixedSizeVariantArray`](https://github.com/gtk-rs/gtk-rs-core/pull/347)
+ * [glib: Mark wrapper structs for `Shared`/`Object`/`Boxed`/`Type` as `#\[repr(transparent)\]`](https://github.com/gtk-rs/gtk-rs-core/pull/348)
+ * [macros: implement FromGlibNone/Full for shared boxed types.](https://github.com/gtk-rs/gtk-rs-core/pull/346)
+ * [Fix #335 - repr(transparent) for cairo\_bool\_t; correct opaque types](https://github.com/gtk-rs/gtk-rs-core/pull/336)
+ * [glib: Implement Send/Sync/PartialOrd/Ord/Hash on GPtrSlice/GSlice](https://github.com/gtk-rs/gtk-rs-core/pull/349)
```
sdroege, line 145:
```markdown
While not used widely yet, a new [`glib::collections`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/collections/index.html) module was added that provides more efficient access to [slices](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/collections/struct.Slice.html), [pointer slices](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/collections/struct.PtrSlice.html) and [linked lists](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/collections/struct.PtrSlice.html) allocated from C. This is similar to the older [`glib::GString`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.GString.html) type, which represents an UTF-8 encoded, `NUL`-terminated C string allocated with the GLib allocator.

Previously (and still for most API), these were converted from/to native Rust types at the FFI boundary, which often caused unnecessary re-allocations and memory copies. In the next release we're trying to move even more API to these new collection types for further performance improvements, and for only doing conversions between Rust and C types when actually necessary.
```

_posts/2022-01-14-new-release.md
sdroege:
```markdown
`glib::String` was renamed to [`glib::GStringBuilder`](https://gtk-rs.org/gtk-rs-core/git/docs/glib/struct.GStringBuilder.html) to avoid confusion with `glib::GString` and to make it clearer that this is meant for building `NUL`-terminated C strings.

In addition the API was simplified considerably:
 * All API that allowed to create invalid UTF-8 was removed, leaving only safe API to build UTF-8 strings.
 * `GStringBuilder` directly dereferences into `&str`, so all `&str` API can be used on it.
 * `GStringBuilder` implements the `std::fmt::Write` trait, which allows making use of various `std` Rust API. For example `write!(&mut builder, "some format string {}", some_value)` can be used directly to write formatted strings into the string builder.
```

_posts/2022-01-14-new-release.md
sdroege: There were also some more changes related to signal emissions IIRC

_posts/2022-01-14-new-release.md
sdroege:
```markdown
All heap-allocated bindings types do now have a memory representation that is equivalent to the underlying raw C pointer. In addition to reducing memory usage compared to before, this also allows for other optimizations and e.g allows to borrow values from C storage in a few places instead of creating copies or increasing their reference count (e.g. in [`glib::PtrSlice`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/collections/struct.PtrSlice.html)).
```

_posts/2022-01-14-new-release.md
sdroege:
```markdown
When spawning non-`Send` [futures](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.MainContext.html#method.spawn_local) or [closures](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.MainContext.html#method.invoke_local), or attaching the main context [channel](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.Receiver.html#method.attach) to the [`glib::MainContext`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.MainContext.html), it is not necessary anymore to have the main context acquired before. This was a common cause of confusion and this constraint was relaxed now. Instead it is only required that that thread default main context is not owned by a different thread yet, and if ownership is moved to a different thread at some point then it will panic at runtime.

Additionally the API for pushing a main context as the [thread default](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.MainContext.html#method.with_thread_default) was refactored to reduce the risk of misusage. Instead of returning a guard value, it can only be used via a closure now to ensure proper nesting of thread default main contexts.

```rust
let ctx = glib::MainContext::new();
ctx.with_thread_default(|| {
    // do something with `ctx` being the thread default main context.
}).expect("Main context already owned by another thread");
```

```

File: _posts/2022-01-14-new-release.md (hunk 11 above)
Comment (sdroege) [reply _posts/2022-01-14-new-release.md:149]: I guess not, it's just a small simplification.

File: _posts/2022-01-14-new-release.md (hunk 12 above)
Comment (sdroege) [reply _posts/2022-01-14-new-release.md:156]:
```markdown
A new [`glib::closure`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/macro.closure.html) and [`glib::closure_local`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/macro.closure_local.html) was added that allows to create a [`glib::Closure`](https://gtk-rs.org/gtk-rs-core/stable/0.16/docs/glib/closure/struct.Closure.html) from a Rust closure while automatically unpacking argument [`glib::Value`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/value/struct.Value.html)s to their native Rust types and packing the return value.

While this doesn't sound too exciting, it makes connecting to signals that have no static bindings more convenient.

Before:

```rust
obj.connect(
    "notify", false,
    |args| {
        let _obj = args[0].get::<glib::Object>().unwrap();
        let pspec = args[1].get::<glib::ParamSpec>().unwrap();
        println!("property notify: {}", pspec.name());
    });
```

After:

```rust
obj.connect_closure(
    "notify", false,
    glib::closure_local!(|_obj: glib::Object, pspec: glib::ParamSpec| {
        println!("property notify: {}", pspec.name());
    }));
```

The macros support the same syntax as the [`glib::clone`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/macro.clone.html) macro for creating strong/weak references, plus a [watch](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/macro.closure.html#object-watching) feature that automatically disconnects the closure once the watched object disappears.

```

File: _posts/2022-01-14-new-release.md (hunk 14 above)
Comment (sdroege) [reply _posts/2022-01-14-new-release.md:182]:
```markdown
[Spawning](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.MainContext.html#method.spawn) futures on the main context now returns the source id. This allows to get access to the underlying [`glib::Source`](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.MainContext.html#method.find_source_by_id) and to [cancel](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/struct.Source.html#method.destroy) the future before completion.

Cancelling the future by this, or by other API that drops it (e.g. by selecting on two futures and only processing the first that resolves), will cancel all asynchronous operations performed by them at the next opportunity.
```

_posts/2022-01-14-new-release.md
sdroege:
```markdown
The GTK3 bindings had API added for generically [setting](https://gtk-rs.org/gtk3-rs/stable/0.15/docs/gtk/prelude/trait.ContainerExtManual.html#tymethod.child_set_property) and [getting](https://gtk-rs.org/gtk3-rs/stable/0.15/docs/gtk/prelude/trait.ContainerExtManual.html#tymethod.child_property) child properties.

This allows to access child properties for which no static bindings are available and performs all checks at runtime instead of at compile time. The API works the same as the API for [setting](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/object/trait.ObjectExt.html#tymethod.set_property) and [getting](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/object/trait.ObjectExt.html#tymethod.property) object properties.
```

_posts/2022-01-14-new-release.md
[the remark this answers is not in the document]
sdroege: # Changes to derive macro names

All [derive](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/index.html#derives) and [attribute](https://gtk-rs.org/gtk-rs-core/stable/0.15/docs/glib/index.html#attributes) macros are named more consistently now. Instead of e.g. `glib::GBoxed`, the new name is just `glib::Boxed` and misses the additional `G`, which would be redundant namespacing.

_posts/2022-01-14-new-release.md
bilelmoussaoui: Similar to the `gtk::StringObject` in gtk4 that wraps a `String` in a `GObject`. You can now use the `glib::BoxedAnyObject` to create a wrapper of your Rust types into a GObject. This is useful for cases where you want to create a `gio::ListModel` of your Rust type without having to manually write the necessary boilerplate to make a GObject wrapper of it.

_posts/2022-01-14-new-release.md
bilelmoussaoui: `glib::ParamSpecOverride` has now two helper constructors to create an override of a class property or an interface property. For example, overriding the `hadjustment` property of the `gtk::Scrollable` interface can be done with `ParamSpecOverride::for_interface::<gtk::Scrollable>("hadjustment")`

_posts/2022-01-14-new-release.md
bilelmoussaoui: - `ObjectExt::property` now return the type T directly instead of a `glib::Value` and panics in case the property is not readable or the wrong type is provided. 
- `ObjectExt::set_property` now panics in case the property is not writeable or doesn't exists
-`ObjectExt::emit_by_name` expects you to pass in the type T for the returned value by the signal. If the signal doesn't return anything, you can use `some_obj.emit_by_name::<()>("some_signal", &[]);`  and also panics if the signal doesn't exists.

A `ObjectExt::try_*` variant exists if you want the old behaviour. As most of the application code unwraps the returned results, this should be one of the slight but nice improvements of the overall API.

sdroege: Can you add 1-2 before/after code examples to make it easier for people to port their code?

bilelmoussaoui: Updated

_posts/2022-01-14-new-release.md
GuillaumeGomez: I think it's simpler to just mention the update of the MSRV if the macros are still backward compatible. :)

PR #282, opened 2022-10-18
[release] merging release-2022-10-18 into master

GuillaumeGomez:
cc @GuillaumeGomez @sdroege @bilelmoussaoui


_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Remove includes](https://github.com/gtk-rs/gtk-rs-core/pull/475)
+ * [CI: publish docs directly to 0.15](https://github.com/gtk-rs/gtk-rs-core/pull/480)
+ * [CI: grab an svg variant of the logo as well](https://github.com/gtk-rs/gtk-rs-core/pull/482)
+ * [CI: grab an svg variant of the logo as well](https://github.com/gtk-rs/gtk-rs-core/pull/481)
+ * [pango: add setters for Rectangle](https://github.com/gtk-rs/gtk-rs-core/pull/486)
+ * [glib: use Result from std for logging macros](https://github.com/gtk-rs/gtk-rs-core/pull/487)
+ * [gio: move task docs on the structs](https://github.com/gtk-rs/gtk-rs-core/pull/488)
+ * [Generate bindings for markup\_escape\_text](https://github.com/gtk-rs/gtk-rs-core/pull/490)
+ * [0.15.2 release](https://github.com/gtk-rs/gtk-rs-core/pull/491)
+ * [implement To/FromVariant for OS strings and paths](https://github.com/gtk-rs/gtk-rs-core/pull/483)
```
sdroege, line 34: Should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Remove redundant allocations](https://github.com/gtk-rs/gtk-rs-core/pull/498)
+ * [pangocairo: don't re-export types on prelude](https://github.com/gtk-rs/gtk-rs-core/pull/499)
+ * [Regenerate with latest gir & simplify windows CI](https://github.com/gtk-rs/gtk-rs-core/pull/496)
+ * [Smaller subclass macro](https://github.com/gtk-rs/gtk-rs-core/pull/500)
+ * [glib: Add an object builder](https://github.com/gtk-rs/gtk-rs-core/pull/503)
```
sdroege, line 39: Should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Update versions to 0.15.1](https://github.com/gtk-rs/gtk-rs-core/pull/476)
+ * [Fix indentation of sys Cargo.toml](https://github.com/gtk-rs/gtk-rs-core/pull/478)
+ * [Update crate versions](https://github.com/gtk-rs/gtk-rs-core/pull/473)
+ * [CI: publish the v0.15 docs](https://github.com/gtk-rs/gtk-rs-core/pull/479)
+ * [Remove includes](https://github.com/gtk-rs/gtk-rs-core/pull/474)
+ * [Remove includes](https://github.com/gtk-rs/gtk-rs-core/pull/475)
+ * [CI: publish docs directly to 0.15](https://github.com/gtk-rs/gtk-rs-core/pull/480)
+ * [CI: grab an svg variant of the logo as well](https://github.com/gtk-rs/gtk-rs-core/pull/482)
+ * [CI: grab an svg variant of the logo as well](https://github.com/gtk-rs/gtk-rs-core/pull/481)
+ * [pango: add setters for Rectangle](https://github.com/gtk-rs/gtk-rs-core/pull/486)
```
sdroege, line 29: Should get an entry, also for all the other boxed-inline types. There are many such PRs here IIRC. @bilelmoussaoui ?

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Generate bindings for markup\_escape\_text](https://github.com/gtk-rs/gtk-rs-core/pull/490)
+ * [0.15.2 release](https://github.com/gtk-rs/gtk-rs-core/pull/491)
+ * [implement To/FromVariant for OS strings and paths](https://github.com/gtk-rs/gtk-rs-core/pull/483)
+ * [Remove redundant allocations](https://github.com/gtk-rs/gtk-rs-core/pull/498)
+ * [pangocairo: don't re-export types on prelude](https://github.com/gtk-rs/gtk-rs-core/pull/499)
+ * [Regenerate with latest gir & simplify windows CI](https://github.com/gtk-rs/gtk-rs-core/pull/496)
+ * [Smaller subclass macro](https://github.com/gtk-rs/gtk-rs-core/pull/500)
+ * [glib: Add an object builder](https://github.com/gtk-rs/gtk-rs-core/pull/503)
+ * [Add more documentation on how to replace g\_signal\_connect\_object](https://github.com/gtk-rs/gtk-rs-core/pull/506)
+ * [glib: Provide `ThreadGuard` and `thread\_id()` as public API](https://github.com/gtk-rs/gtk-rs-core/pull/504)
```
sdroege, line 41: Should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [0.15 backports](https://github.com/gtk-rs/gtk-rs-core/pull/510)
+ * [Don't require `Send` closures for GIO-style async functions](https://github.com/gtk-rs/gtk-rs-core/pull/511)
+ * [gio: Don't implement `Send` on `GioFuture` anymore](https://github.com/gtk-rs/gtk-rs-core/pull/512)
```
sdroege, line 44: These two should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Update gir-files](https://github.com/gtk-rs/gtk-rs-core/pull/513)
+ * [gio: add Initable::with\_type](https://github.com/gtk-rs/gtk-rs-core/pull/515)
```
sdroege, line 46: Should get an entry, together with the other (async) initable changes

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [translate: Pre-allocate Rust hashmap at correct size](https://github.com/gtk-rs/gtk-rs-core/pull/561)
+ * [0.15 backports](https://github.com/gtk-rs/gtk-rs-core/pull/562)
+ * [glib-macros: Pass a pointer for watched objects into closure macro](https://github.com/gtk-rs/gtk-rs-core/pull/563)
+ * [0.15 backports](https://github.com/gtk-rs/gtk-rs-core/pull/566)
+ * [Variant::from\_dict\_entry](https://github.com/gtk-rs/gtk-rs-core/pull/564)
+ * [Smaller spec numeric definition](https://github.com/gtk-rs/gtk-rs-core/pull/568)
+ * [0.15 backports](https://github.com/gtk-rs/gtk-rs-core/pull/570)
+ * [pango: Add LayoutLine.runs()](https://github.com/gtk-rs/gtk-rs-core/pull/571)
+ * [pango: Add missing getter for GlyphItemIter.](https://github.com/gtk-rs/gtk-rs-core/pull/572)
+ * [define a corresponding `Builder` for each `ParamSpec`](https://github.com/gtk-rs/gtk-rs-core/pull/551)
```
sdroege, line 77: Should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [glib: impl AsRef&lt;Variant&gt; for Variant](https://github.com/gtk-rs/gtk-rs-core/pull/565)
+ * [glib-macros: Add support for enums to glib::Variant ](https://github.com/gtk-rs/gtk-rs-core/pull/434)
```
sdroege, line 79: Should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [use Self where possible in wrapper macros](https://github.com/gtk-rs/gtk-rs-core/pull/575)
+ * [remove problematic phantom in glib\_shared\_wrapper](https://github.com/gtk-rs/gtk-rs-core/pull/569)
+ * [Fix clippy lints](https://github.com/gtk-rs/gtk-rs-core/pull/581)
+ * [Fix old/broken gnome/gtk doc links](https://github.com/gtk-rs/gtk-rs-core/pull/582)
+ * [glib: Use "%s" instead of replacing % in log functions](https://github.com/gtk-rs/gtk-rs-core/pull/583)
+ * [impl `ToValue`, `FromValue` for char](https://github.com/gtk-rs/gtk-rs-core/pull/579)
+ * [glib: Correctly cast log messages to C char types](https://github.com/gtk-rs/gtk-rs-core/pull/584)
+ * [glib: Automatically implement `Send + Sync` on object subclasses if t…](https://github.com/gtk-rs/gtk-rs-core/pull/586)
```
sdroege, line 87: Should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Use glib 2.71 for minimum v2\_72 pkg-config versions](https://github.com/gtk-rs/gtk-rs-core/pull/587)
+ * [glib: Add function to wrap a future with a timeout](https://github.com/gtk-rs/gtk-rs-core/pull/528)
```
sdroege, line 89: Should get an entry, and also mention that there are more futures utility things are planned for the future

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [glib: Allow using other error types and value type checkers for …](https://github.com/gtk-rs/gtk-rs-core/pull/588)
+ * [glib: Only auto-impl `Send+Sync` for subclasses if the parent type does](https://github.com/gtk-rs/gtk-rs-core/pull/591)
+ * [glib: Use the correct value type checker for optional types](https://github.com/gtk-rs/gtk-rs-core/pull/592)
+ * [Add new classes from glib 2.72](https://github.com/gtk-rs/gtk-rs-core/pull/577)
```
sdroege, line 93: Should get an entry for the new versions of things we support, also GTK4 @bilelmoussaoui

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [glib-macros: Fix clippy lints](https://github.com/gtk-rs/gtk-rs-core/pull/614)
+ * [glib: Fix build with `v2\_72` and require 2.72.0 now](https://github.com/gtk-rs/gtk-rs-core/pull/615)
+ * [Fix a couple of new clippy warnings](https://github.com/gtk-rs/gtk-rs-core/pull/616)
+ * [glib: Generate docs for BindingGroupBuilder](https://github.com/gtk-rs/gtk-rs-core/pull/618)
+ * [Update gir](https://github.com/gtk-rs/gtk-rs-core/pull/619)
+ * [More Cairo GValue fixes](https://github.com/gtk-rs/gtk-rs-core/pull/623)
+ * [Implement ValueTypeOptional for BoxedInline and cairo types](https://github.com/gtk-rs/gtk-rs-core/pull/621)
+ * [glib: implement ValueTypeOptional for Variant, VariantType, BoxedValue](https://github.com/gtk-rs/gtk-rs-core/pull/626)
+ * [Add rustfmt.toml to allow some editors to auto-format the code on save](https://github.com/gtk-rs/gtk-rs-core/pull/627)
+ * [Update minimum supported GLib version to 2.56](https://github.com/gtk-rs/gtk-rs-core/pull/631)
```
sdroege, line 119: also this for minimum supported versions

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [glib: Fix `mkdtemp()` and `mkdtemp\_full()`](https://github.com/gtk-rs/gtk-rs-core/pull/630)
+ * [glib/variant: add some more safe wrappers](https://github.com/gtk-rs/gtk-rs-core/pull/632)
+ * [glib: Remove ending NUL when converting Variant to OsString.](https://github.com/gtk-rs/gtk-rs-core/pull/625)
+ * [error: add `matches` method](https://github.com/gtk-rs/gtk-rs-core/pull/633)
+ * [Fix Windows build](https://github.com/gtk-rs/gtk-rs-core/pull/634)
+ * [glib: Remove SendUnique and SendUniqueCell](https://github.com/gtk-rs/gtk-rs-core/pull/635)
+ * [glib: Add new `ObjectImplRef` and `ObjectImplWeakRef` types](https://github.com/gtk-rs/gtk-rs-core/pull/636)
```
sdroege, line 126: Should get an entry, together with `BorrowedObject` and the `ToOwned` etc parts. Also together with the removal of `Self::Type`

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Remove more redundant null checks](https://github.com/gtk-rs/gtk-rs-core/pull/648)
+ * [glib: Don't serialize 0 value flags as variants](https://github.com/gtk-rs/gtk-rs-core/pull/624)
+ * [glib: Add bindings for GSignalGroup](https://github.com/gtk-rs/gtk-rs-core/pull/644)
+ * [gio: Add local variant for connect\_cancelled](https://github.com/gtk-rs/gtk-rs-core/pull/650)
+ * [glib: WatchedObject improvements](https://github.com/gtk-rs/gtk-rs-core/pull/651)
+ * [glib-macros: Remove boxed\_nullable and shared\_boxed\_nullable attribute for glib::Boxed](https://github.com/gtk-rs/gtk-rs-core/pull/653)
+ * [glib: Add getter for `glib::Error` domain quark](https://github.com/gtk-rs/gtk-rs-core/pull/655)
+ * [0.15 backports](https://github.com/gtk-rs/gtk-rs-core/pull/657)
+ * [Set properties improvements](https://github.com/gtk-rs/gtk-rs-core/pull/660)
+ * [impl fallible IntoIterator for ListModel,ListStore](https://github.com/gtk-rs/gtk-rs-core/pull/668)
```
sdroege, line 141: Should get an entry maybe? How useful is it, people asked often for it but I never needed it @bilelmoussaoui ?

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Update gir files](https://github.com/gtk-rs/gtk-rs-core/pull/667)
+ * [cairo+gio+glib: Add safety docs for low hanging fruits](https://github.com/gtk-rs/gtk-rs-core/pull/671)
+ * [ glib: Add WeakRefNotify (v2)](https://github.com/gtk-rs/gtk-rs-core/pull/666)
+ * [Move `gio::compile\_resources` to its own crate](https://github.com/gtk-rs/gtk-rs-core/pull/670)
```
sdroege, line 145: Should get an entry, new `glib-build-tools` crate. Also related to the book @Hofer-Julian

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Add IntoGlibPtr trait](https://github.com/gtk-rs/gtk-rs-core/pull/673)
```
sdroege, line 146: Should get an entry, with a hint that we'll make use more of that in the future to reduce unnecessary copies when owned values are given away to C

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [cairo: Add missing User Fonts](https://github.com/gtk-rs/gtk-rs-core/pull/672)
+ * [glib/gio: Implement `FusedFuture` / `FusedStream` in some places](https://github.com/gtk-rs/gtk-rs-core/pull/677)
+ * [Port Dockerfile to fedora rawhide and use system deps from git](https://github.com/gtk-rs/gtk-rs-core/pull/676)
+ * [Implement `FusedIterator` for various custom iterators](https://github.com/gtk-rs/gtk-rs-core/pull/678)
+ * [cairo: Fix user font callbacks.](https://github.com/gtk-rs/gtk-rs-core/pull/679)
+ * [CI: update the gh-package job](https://github.com/gtk-rs/gtk-rs-core/pull/681)
+ * [Add some more tests for various costum `Iterator` impls](https://github.com/gtk-rs/gtk-rs-core/pull/680)
+ * [Don't build gdk-pixbuf tests in the docker image](https://github.com/gtk-rs/gtk-rs-core/pull/682)
+ * [gio: add AsyncInitable subclassing support](https://github.com/gtk-rs/gtk-rs-core/pull/654)
+ * [gio: Use guard objects for `Application::hold()` and `mark\_busy()`](https://github.com/gtk-rs/gtk-rs-core/pull/685)
```
sdroege, line 157: Should get an entry

_posts/2022-10-18-new-release.md
Hofer-Julian: The book is already adapted if that's what you mean :)

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [glib: Fix `ParamSpec::name()` return value lifetime](https://github.com/gtk-rs/gtk-rs-core/pull/728)
+ * [glib: fix ParamSpec with null blurb and nick](https://github.com/gtk-rs/gtk-rs-core/pull/726)
+ * [pango: Mark parameters to `extents\_to\_pixels()` as mutable](https://github.com/gtk-rs/gtk-rs-core/pull/730)
+ * [add set\_glyph to GlyphInfo](https://github.com/gtk-rs/gtk-rs-core/pull/736)
+ * [Regenerate](https://github.com/gtk-rs/gtk-rs-core/pull/717)
+ * [fix glyph string methods that don't need to be &mut](https://github.com/gtk-rs/gtk-rs-core/pull/737)
+ * [glib: add doc alias for g\_type\_class\_ref](https://github.com/gtk-rs/gtk-rs-core/pull/738)
+ * [glib: bind iconv functions](https://github.com/gtk-rs/gtk-rs-core/pull/708)
+ * [cairo: Update to freetype 0.31](https://github.com/gtk-rs/gtk-rs-core/pull/741)
+ * [glib: Add flag setters to ParamSpec builders](https://github.com/gtk-rs/gtk-rs-core/pull/735)
```
sdroege, line 196: Should get an entry, together with the builders in general

sdroege: Also `SignalBuilder` is new or not?

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [glib: Various paramspec builder API improvements](https://github.com/gtk-rs/gtk-rs-core/pull/761)
+ * [glib/gio: Replace various type parameters with `impl trait`](https://github.com/gtk-rs/gtk-rs-core/pull/762)
+ * [glib/types: use rustc-hash for HashMap](https://github.com/gtk-rs/gtk-rs-core/pull/755)
+ * [glib/subclass/types: use BTreeMap instead of HashMap](https://github.com/gtk-rs/gtk-rs-core/pull/764)
+ * [gio: make Task::cancellable return Option](https://github.com/gtk-rs/gtk-rs-core/pull/767)
+ * [Implement Borrow on object types](https://github.com/gtk-rs/gtk-rs-core/pull/713)
+ * [glib: Reduce number of allocations when writing non-structured logs v…](https://github.com/gtk-rs/gtk-rs-core/pull/770)
+ * [gio: `CancellableFuture` ergonomic for cancelling tasks](https://github.com/gtk-rs/gtk-rs-core/pull/772)
+ * [glib: Don't allow passing a construct(-only) property twice during ob…](https://github.com/gtk-rs/gtk-rs-core/pull/774)
+ * [glib: Panic in Object::new() and related functions and add try\_new()](https://github.com/gtk-rs/gtk-rs-core/pull/771)
```
sdroege, line 219: Should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Update gir](https://github.com/gtk-rs/gtk-rs-core/pull/759)
+ * [Don't checkout submodules by default](https://github.com/gtk-rs/gtk-rs-core/pull/760)
+ * [glib: Various paramspec builder API improvements](https://github.com/gtk-rs/gtk-rs-core/pull/761)
+ * [glib/gio: Replace various type parameters with `impl trait`](https://github.com/gtk-rs/gtk-rs-core/pull/762)
+ * [glib/types: use rustc-hash for HashMap](https://github.com/gtk-rs/gtk-rs-core/pull/755)
+ * [glib/subclass/types: use BTreeMap instead of HashMap](https://github.com/gtk-rs/gtk-rs-core/pull/764)
+ * [gio: make Task::cancellable return Option](https://github.com/gtk-rs/gtk-rs-core/pull/767)
+ * [Implement Borrow on object types](https://github.com/gtk-rs/gtk-rs-core/pull/713)
+ * [glib: Reduce number of allocations when writing non-structured logs v…](https://github.com/gtk-rs/gtk-rs-core/pull/770)
+ * [gio: `CancellableFuture` ergonomic for cancelling tasks](https://github.com/gtk-rs/gtk-rs-core/pull/772)
```
sdroege, line 217: Should get an entry, together with the timeout future thing above

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Add `#\[must\_use\]` to guard objects](https://github.com/gtk-rs/gtk-rs-core/pull/752)
+ * [glib/object: add track\_caller annotation in a few places](https://github.com/gtk-rs/gtk-rs-core/pull/754)
+ * [glib: Remove unneeded and wrong `upcast()` from `BoxedAnyObject` docs…](https://github.com/gtk-rs/gtk-rs-core/pull/758)
+ * [Update gir](https://github.com/gtk-rs/gtk-rs-core/pull/759)
+ * [Don't checkout submodules by default](https://github.com/gtk-rs/gtk-rs-core/pull/760)
+ * [glib: Various paramspec builder API improvements](https://github.com/gtk-rs/gtk-rs-core/pull/761)
+ * [glib/gio: Replace various type parameters with `impl trait`](https://github.com/gtk-rs/gtk-rs-core/pull/762)
+ * [glib/types: use rustc-hash for HashMap](https://github.com/gtk-rs/gtk-rs-core/pull/755)
+ * [glib/subclass/types: use BTreeMap instead of HashMap](https://github.com/gtk-rs/gtk-rs-core/pull/764)
+ * [gio: make Task::cancellable return Option](https://github.com/gtk-rs/gtk-rs-core/pull/767)
```
sdroege, line 214: Should get an entry, task subclassing bindings etc

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [Implement Borrow on object types](https://github.com/gtk-rs/gtk-rs-core/pull/713)
+ * [glib: Reduce number of allocations when writing non-structured logs v…](https://github.com/gtk-rs/gtk-rs-core/pull/770)
+ * [gio: `CancellableFuture` ergonomic for cancelling tasks](https://github.com/gtk-rs/gtk-rs-core/pull/772)
+ * [glib: Don't allow passing a construct(-only) property twice during ob…](https://github.com/gtk-rs/gtk-rs-core/pull/774)
+ * [glib: Panic in Object::new() and related functions and add try\_new()](https://github.com/gtk-rs/gtk-rs-core/pull/771)
+ * [Implement more FFI translation traits for shared types](https://github.com/gtk-rs/gtk-rs-core/pull/756)
+ * [cairo: Use AsRef&lt;T&gt; for custom subclassing types](https://github.com/gtk-rs/gtk-rs-core/pull/786)
+ * [glib/BindingBuilder: add concrete type based transform setters](https://github.com/gtk-rs/gtk-rs-core/pull/782)
```
sdroege, line 222: Should get an entry

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [pango: Manually implement GlyphItemIter and AttrIterator with lifetimes](https://github.com/gtk-rs/gtk-rs-core/pull/784)
+ * [Pango odds and ends](https://github.com/gtk-rs/gtk-rs-core/pull/785)
+ * [Change \*Impl trait methods to only take &self and not Self::Type in addition](https://github.com/gtk-rs/gtk-rs-core/pull/783)
```
sdroege, line 225: Should get an entry

_posts/2022-10-18-new-release.md
sdroege: The book is already updated to gtk4 0.5.0? Good, then can link to that :)

_posts/2022-10-18-new-release.md
bilelmoussaoui: > Also `SignalBuilder` is new or not?

no it is not, but had some api changes

_posts/2022-10-18-new-release.md
```diff
@@ -0,0 +1,546 @@
...
+ * [pango: Mark parameters to `extents\_to\_pixels()` as mutable](https://github.com/gtk-rs/gtk-rs-core/pull/730)
+ * [add set\_glyph to GlyphInfo](https://github.com/gtk-rs/gtk-rs-core/pull/736)
+ * [Regenerate](https://github.com/gtk-rs/gtk-rs-core/pull/717)
+ * [fix glyph string methods that don't need to be &mut](https://github.com/gtk-rs/gtk-rs-core/pull/737)
+ * [glib: add doc alias for g\_type\_class\_ref](https://github.com/gtk-rs/gtk-rs-core/pull/738)
+ * [glib: bind iconv functions](https://github.com/gtk-rs/gtk-rs-core/pull/708)
+ * [cairo: Update to freetype 0.31](https://github.com/gtk-rs/gtk-rs-core/pull/741)
+ * [glib: Add flag setters to ParamSpec builders](https://github.com/gtk-rs/gtk-rs-core/pull/735)
+ * [Fix new clippy warnings](https://github.com/gtk-rs/gtk-rs-core/pull/742)
+ * [gio: Manually implement ActionEntry](https://github.com/gtk-rs/gtk-rs-core/pull/743)
```
bilelmoussaoui, line 198: Should get an entry

bilelmoussaoui:
For gtk4


### GTK 4

Along with plenty of bugfixes and small improvements, the 0.5 release of gtk4-rs brings a couple of useful features

- `gsk::Transform`

Most of the C functions return `NULL` which represents an identity transformation. The Rust API nowadays returns an identity 
transformation instead of `None`

- `#[gtk::CompositeTemplate]` runtime validation

If used with `#[template(string=...)]` or `#[template(file=...)]` and the `xml_validation` feature is enabled, the XML content will be validated to ensure the child widgets that the code is trying to retrieve exists.

- `#[gtk::test]`

As GTK is single-threaded, using it in your tests is problematic as you can't initialize GTK multiple times. The new attribute macro helps with that as it runs the tests in a thread pool instead of in parallel.

Details at <https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4_macros/attr.test.html>

- `WidgetClassSubclassExt::install_action_async`

In a lot of use cases, people used to do 

```rust
klass.install_action("some_group.action", None, |widget, _, _| {
    let ctx = glib::MainContext::default();
    ctx.spawn_local(clone!(@weak widget => async move {
        /// call some async function
        widget.async_function().await;
    }));
})
```

The new helper function allows you to write the code above like

```rust
klass.install_action_async("some_group.action", None, |widget, _, _| async move {
    /// call some async function
    widget.async_function().await;
})
```
Which helps avoiding the usage of the `clone` macro in some cases.

- GTK 4.8 API additions which can be enabled with `v4_8`

- Various fixes for the gdk-wayland & wayland-rs integration requiring wayland-client v0.30

- Plenty of new examples:
    * Gif Paintable for rendering Gifs
    * ColumnView for displaying data in a table-like format
    * Confetti animation
    * Rotation / Squeezer example

_posts/2022-10-18-new-release.md
sdroege: Previously when creating a GObject subclass, all virtual methods passed the implementation struct and the object itself as arguments, e.g.

```rust
pub trait ObjectImpl {
    fn constructed(&self, obj: &Self::Type);
}
```

This caused a lot of confusion and was also redudant. Instead, starting with this release only the implementation struct is passed by reference, e.g.

```rust
pub trait ObjectImpl {
    fn constructed(&self);
}
```

In most contexts the object itself was not actually needed, so this also simplifies the code. For the cases when the object is needed, it can be retrieved via the `instance()` method on the implementation struct

```rust
impl ObjectImpl for MyObject {
    fn constructed(&self) {
        let instance = self.instance();
        instance.do_something();
    }
}
```

In a similar way it is also possible to retrieve the implementation struct from the instance via the `imp()` method. Both methods are cheap and only involve some pointer arithmetic.

Additionally, to make it easy to pass around the implementation struct into e.g. closures, there is now also a reference counted wrapper around it available (`ObjectImplRef`) that can be retrieved via `imp.to_owned()` or `imp.ref_counted()`, and a weak reference variant that can be retrieved via `imp.downgrade()`. Both are working in combination with the `glib::clone!` macro, too.

```rust
impl ObjectImpl for MyObject {
    fn constructed(&self) {
        // for a strong reference
        self.button.connect_clicked(glib::clone!(@to-owned self as imp => move |button| {
            imp.do_something();
        }));

        // for a weak reference
        self.button.connect_clicked(glib::clone!(@weak self as imp => move |button| {
            imp.do_something();
        }));
    }
}
```

_posts/2022-10-18-new-release.md
sdroege: See https://github.com/gtk-rs/gtk-rs.github.io/pull/282#discussion_r1001698249

_posts/2022-10-18-new-release.md
sdroege: You write about the signal/pspec builders, with some example code?

_posts/2022-10-18-new-release.md
sdroege: Object property bindings allow for transform functions to be defined, which convert the property value to something else for the other object whenever it changes. Previously these were defined on the generic `glib::Value` type, but as the types are generally fixed and known in advance it is now possible to define them directly with the correct types.

```rust
        source
            .bind_property("name", &target, "name")
            .flags(crate::BindingFlags::SYNC_CREATE)
            .transform_to(|_binding, value: &str| Some(format!("{} World", value)))
            .transform_from(|_binding, value: &str| Some(format!("{} World", value)))
            .build();
```

If the types don't match then this is considered a programming error and will panic at runtime.

The old way of defining transform functions via `glib::Value` is still available via new functions

```rust
        source
            .bind_property("name", &target, "name")
            .flags(crate::BindingFlags::SYNC_CREATE)
            .transform_to_with_values(|_binding, value| {
                let value = value.get::<&str>().unwrap();
                Some(format!("{} World", value).to_value())
            })
            })
            .transform_from_with_values(|_binding, value| {
                let value = value.get::<&str>().unwrap();
                Some(format!("{} World", value).to_value())
            })
            .build();
```

_posts/2022-10-18-new-release.md
sdroege: `glib::Object::new()` returned a `Result` in previous versions. However, the only way how this could potentially fail was via a programming error: properties that don't exist were tried to be passed, or values of the wrong type were set for a property. By returning a `Result`, the impression was given that this can also fail in a normal way that is supposed to be handled by the caller.

As this is not the case, `glib::Object::new()` always panics if the arguments passed to it are invalid and no longer returns a `Result`.

In the same spirit, `object.try_set_property()`, `object.try_property()`, `object.try_emit()` and `object.try_connect()` do not exist any longer and only the panicking variants are left as the only way they could fail was if invalid arguments are provided.

_posts/2022-10-18-new-release.md
bilelmoussaoui: I have added an entry for gtk4 in the gtk4 section already

_posts/2022-10-18-new-release.md
sdroege: The minimum supported version of the C library of various crates was updated to the versions available in Ubuntu 18.04:

  * GLib/GIO requires at least version 2.56 now and supports API up to version 2.74
  * gdk-pixbuf requires at least version 2.36.8 and supports API up to version 2.42
  * Pango requires at least version 1.40 and supports API up to version 1.52
  * GTK3 requires at least version 3.22.30 and supports API up to version 3.24.30
  * GTK4 requires at least version 4.0.0 and supports API up to version 4.10

The minimum supported Rust version by all crates was updated to version 1.63.

CC @bilelmoussaoui please confirm

_posts/2022-10-18-new-release.md
sdroege: In addition to construction objects dynamically, or e.g. when implementing new GObject subclasses, via `glib::Object::new()` and related API, there is also a more convenient object builder available.

```rust
let obj = glib::Object::builder::<MyObject>()
    .property("prop1", 123)
    .property("prop2", "test")
    .property("prop3", false)
    .build();
```

This allows for slightly more idiomatic Rust code.

_posts/2022-10-18-new-release.md
bilelmoussaoui: it is correct for gtk4 yes, but I wouldn't mention v4.10 as it is not released yet

_posts/2022-10-18-new-release.md
sdroege: Asynchronous operations on GIO objects previously required the completion closure to be `Send`. This is not required anymore as the objects themselves are not `Send`-able either and the operation will be completed on the current thread via the thread's own main context. This should make usage of the asynchronous operations easier from GTK applications, where all UI objects are not `Send`.

As a result of this, also futures provided by GIO objects based on these asynchronous operations do not implement the `Send` trait anymore, which was wrong to begin with and caused panics at runtime in any situation where the future was actually used on different threads.

_posts/2022-10-18-new-release.md
sdroege: Actually I don't really know what to write about this. That seems @pbor territory :)

_posts/2022-10-18-new-release.md
sdroege: Starting with the previous release it was possible to derive the required trait implementations for (de)serializing Rust structs from/to `glib::Variant`s. With this release support for enums is also added, with different options for how the enum is going to be represented. Both C-style enums as well as enums with values in the variants are supported.

```rust
#[derive(Debug, PartialEq, Eq, glib::Variant)]
enum Foo {
    MyA,
    MyB(i32),
    MyC { some_int: u32, some_string: String }
}
let v = Foo::MyC { some_int: 1, some_string: String::from("bar") };
let var = v.to_variant();
assert_eq!(var.child_value(0).str(), Some("my-c"));
assert_eq!(var.get::<Foo>(), Some(v));
```

_posts/2022-10-18-new-release.md
sdroege: A couple of futures helper functions were added with this release that should make futures usage easier in GTK applications.

 * Timeouts: `glib::future_with_timeout(Duration::from_millis(20), fut)` will resolve to the result of `fut` if it takes less than 20ms or otherwise to an error
 * Cancellation: `gio::CancellableFuture::new(fut, cancellable)` will resolve to the future unless the `cancellable` is cancelled before that

_posts/2022-10-18-new-release.md
sdroege: Maybe not because nothing is really implemented yet for that

_posts/2022-10-18-new-release.md
sdroege: Also @pbor territory, I don't know what to write here

_posts/2022-10-18-new-release.md
sdroege: I'll leave this one to you @bilelmoussaoui :)

_posts/2022-10-18-new-release.md
Hofer-Julian: Okay!
https://github.com/gtk-rs/gtk4-rs/pull/1024

_posts/2022-10-18-new-release.md
sdroege: When creating properties for GObject subclasses they need to be declared beforehand via a `glib::ParamSpec`. Previously these had simple `new()` functions with lots of parameters. These still exist but it's usually more convenient to use the builder pattern to construct them, especially as most of the parameters have sensible defaults.

```rust
// Now
let pspec = glib::ParamSpecUInt("name")
    .maximum(1000)
    .construct()
    .build();

// Previously
let pspec = glib::ParamSpecUInt::new("name", None, None, 0, 1000, 0, glib::ParamFlags::READWRITE | glib::ParamFlags::CONSTRUCT);
```

In a similar spirit, signal definitions are available via a builder. This was available in the previous already but usage was simplified, for example by defaulting to no signal arguments and `()` signal return type.

```rust
// Now
let signal = glib::subclass::Signal::builder("closed").build();

// Before
let signal = glib::subclass::Signal::builder("closed", &[], glib::Type::UNIT).build();
```

_posts/2022-10-18-new-release.md
bilelmoussaoui: I never needed it myself neither, I would just skip

_posts/2022-10-18-new-release.md
bilelmoussaoui: It is now possible to use `gio::ActionEntryBuilder` to construct a `gio::SimpleAction`, the advantage of using that is the `gio::ActionMap` type is passed as a first parameter to the `activate` callback and so avoids the usage of the `clone!` macro.

Before:
```rust
let action = gio::SimpleAction::new("some_action", None);
action.connect_activate(clone!(@weak some_obj => move |_, _| {
 /// do something
});
actions_group.add_action(&action);
```

After
```rust
let action = gio::ActionEntry::builder("some_action").activate(move |some_obj: &SomeType, _, _| {
/// do something
}).build();
actions_group.add_action_entries([action]).unwrap(); // it is safe to unwrap as we don't pass any parameter type that requires validation
```

_posts/2022-10-18-new-release.md
pbor: this is not really newsworthy, just returning Option where we should... clearly no one was using this api :)

PR #326, opened 2025-10-14
Made each project's description visible when hovering over it…

LucasDondo:
… for easier understanding of what the project is for.

### DIFF
--- a/_includes/projects.html
+++ b/_includes/projects.html
@@ -4,7 +4,7 @@
 
 <ul class="projects-overview">
 {% for project in site.data.projects %}
-  <li>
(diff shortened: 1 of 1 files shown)


sdroege:
@sophie-h @bilelmoussaoui Opinions?

LucasDondo:
In case there are doubts because of the inconsistency that not every project has a description, I was planning on a PR adding the description of every project (in case this one gets merged).

LucasDondo:
Also, note that the descriptions were taken from the app's website/app store. So, these are **not** descriptions made up by me or any AI 😆.

